PR #139, opened 2021-02-22
SystemInfo -> ClusterStore and Security Options should be optional?

N0K0:
Hi! Thanks for this nice wrapper!
While testing this tool on some of my daemons i noticed that two fields from docker system info is often missing.

Namely the Security Options and Cluster Store 🤔

Funnily enough both are missing from the output example in their docs too.
https://docs.docker.com/engine/reference/commandline/info/

Seems like both are only printed if there is any content
https://github.com/moby/moby/blob/6d05bba74baf29523187f290a482875f3139e2b9/daemon/info.go#L180-L206
https://github.com/moby/moby/blob/6d05bba74baf29523187f290a482875f3139e2b9/daemon/info.go#L135-L143

### DIFF
--- a/python_on_whales/components/system.py
+++ b/python_on_whales/components/system.py
@@ -217,7 +217,7 @@ class SystemInfo(DockerCamelModel):
     labels: Dict[str, str]
     experimental_build: bool
     server_version: str
-    cluster_store: str
+    cluster_store: Optional[str]
     runtimes: Dict[str, Runtime]
     default_runtime: str
     swarm: SwarmInfo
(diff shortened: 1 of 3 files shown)


gabrieldemarmiesse:
Thank you for the PR, that's great! Could you add a test for this? The easiest way is to take a json that python-on-whales had issues with and to drop it in this directory; https://github.com/gabrieldemarmiesse/python-on-whales/tree/master/tests/python_on_whales/components/jsons/system_info the test suite will automatically pick it up and try to load it.

N0K0:
@gabrieldemarmiesse Done!

gabrieldemarmiesse:
Thanks!

gabrieldemarmiesse (approved):

PR #158, opened 2021-04-18
Allow access to local environment variables when running buildx commands

### LINKED ISSUE #157: ssh argument not working as expected
I have been working on porting a bash script which I use to build some of my docker images to python, and your package has been a great help! I appreciate how easy and straightforward it is to port my docker run commands directly to python.

I just ran into an issue with the `ssh` arg though. Looks like when setting `ssh=default` in the `docker.build()` command, I see an error:
```
error: invalid empty ssh agent socket, make sure SSH_AUTH_SOCK is set
```

I checked, and that variable is set in my shell environment. Any suggestions? Am I using the argument incorrectly? The same bash script with `--ssh default` is working without problems.

Let me know if I can provide anything to help debug. Thanks!

```
    raise DockerException(
python_on_whales.utils.DockerException: The docker command executed was `/usr/bin/docker buildx build --build-arg FROM_TAG=gui-support --ssh default --file docker/ros/kinetic/autoyard-core/Dockerfile --tag ros-kinetic-autoyard-core:test --iidfile /tmp/tmpc9q55l8w/id_file.txt .`.
It returned with code 1

```

Logged the variable with `print("SSH socket {}".format(os.environ['SSH_AUTH_SOCK']))` and can see that the variable is printed correctly, but the docker build command still fails:
```
SSH socket /run/user/1001/keyring/ssh
error: invalid empty ssh agent socket, make sure SSH_AUTH_SOCK is set
```

heemonsu:
Attempts to fix #157

### DIFF
--- a/python_on_whales/utils.py
+++ b/python_on_whales/utils.py
@@ -1,3 +1,4 @@
+import os
 import subprocess
 import warnings
 from pathlib import Path
@@ -104,7 +105,8 @@ def run(
     args = [str(x) for x in args]
     if args[1] == "buildx":
         install_buildx_if_needed(args[0])
-        env["DOCKER_CLI_EXPERIMENTAL"] = "enabled"
+        os.environ["DOCKER_CLI_EXPERIMENTAL"] = "enabled"
+        env = None
     if env == {}:
         env = None
     if capture_stdout:


gabrieldemarmiesse:
As said in #157, we need some other fix to make it work in every case, not just buildx. I'll send a fix soon.

PR #165, opened 2021-05-10
Change the connection mode to AWS ECR using default AWS library

anouvel:
This PR aims to use the default AWS login strategy (embedded in botocore) to ease the use of ECR. Here is an example:

```
# replace by your profile
❯ export AWS_PROFILE=aws_named_profile
❯ python -c "from python_on_whales import docker ; docker.login_ecr()"
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
Login Succeeded
```

It also supports AWS_ACCESS_KEY_ID and all other settings described [here](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html).

Another benefit: it allows to overload settings such as AWS endpoints etc.

## Deprecation

It deprecates the optional parameters of the `login_ecr` method (access key and secret key as argument is not supported anymore).

### DIFF
--- a/python_on_whales/docker_client.py
+++ b/python_on_whales/docker_client.py
@@ -200,6 +200,8 @@ class DockerClient(DockerCLICaller):
         ```python
         botocore.session.get_session().create_client(...)
         ```
+
+        You need botocore to run this function. Use `pip install botocore` to install it.
         """
         import botocore.session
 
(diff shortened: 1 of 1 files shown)


pySilver:
Oh thats nice! Hopefully it will be merged soon!

gabrieldemarmiesse:
I was super busy last week so I didn't have much time to take care of it. I'll modify this PR a bit to make sure we're backward compatible. You can expect a release aroud thursday

PR #182, opened 2021-06-10
Add filters argument for image.list()

haruishi43:
- Added a simple `filters=dict` argument to your `image.list()` function.
- It follows the same format as your `container.list()` function.

I didn't write the test codes for it since I haven't really studied your test codes yet. If you could tell me what is the best approach, I would implement it. I was thinking on the line of using `docker.build(..., labels={"foo": "bar", ...}, ...)` and listing out the images using `docker.image.list(fitlers={"label": "foo=bar"})` and checking for the number to match. I looked into `test_image.py`, but couldn't fully understand how you would go about in an optimal way.

### DIFF
--- a/python_on_whales/components/image/cli_wrapper.py
+++ b/python_on_whales/components/image/cli_wrapper.py
@@ -289,7 +289,7 @@ class ImageCLI(DockerCLICaller):
             raise DockerException(full_cmd, exit_code)
         return stdout.decode().splitlines()
 
-    def list(self) -> List[Image]:
+    def list(self, filters: Dict[str, str] = {}) -> List[Image]:
         """Returns the list of Docker images present on the machine.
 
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse:
Thank you for the pull request! I'm on holidays right now but I'll take a look during the next days. I think it's a good idea to test using the labels but a simpler way could be to pull some images and filter by name. I'll take a look this week-end probably :)

haruishi43:
@gabrieldemarmiesse Looking forward to it! Your project has been a lot of help, thank you so much!

gabrieldemarmiesse (approved):

gabrieldemarmiesse:
I added a very small test, but it's enough to verify that the CLI invocation has no issues

PR #192, opened 2021-07-01
Added py.typed

mdantonio:


gabrieldemarmiesse (approved):
I hope it will be enough! I'll make a release so that you can test and report if it works :)

PR #214, opened 2021-07-15
[buildx] Allow multiple cache_(from|to) directives

redref:
Buildx support multiple directives for cache-from and cache-to.

Extended the wrapper to do so.

### DIFF
--- a/tests/python_on_whales/components/buildx/test_buildx_cli_wrapper.py
+++ b/tests/python_on_whales/components/buildx/test_buildx_cli_wrapper.py
@@ -103,6 +103,19 @@ def test_buildx_caching_dict_form(tmp_path):
 
 
 @pytest.mark.usefixtures("with_docker_driver")
+def test_buildx_caching_list_form(tmp_path):
+    (tmp_path / "Dockerfile").write_text(dockerfile_content1)
+    with docker.buildx.create(driver_options=dict(network="host"), use=True):
+        docker.buildx.build(
+            tmp_path,
+            cache_from=[
+                dict(type="local", src=tmp_path / "cache"),
+                dict(type="local", src=tmp_path / "cache2"),
+            ],
+        )
+
+
+@pytest.mark.usefixtures("with_docker_driver")
 def test_buildx_build_output_type_registry(tmp_path, docker_registry):
     (tmp_path / "Dockerfile").write_text(dockerfile_content1)
     with docker.buildx.create(use=True, driver_options=dict(network="host")):
--- a/python_on_whales/components/buildx/cli_wrapper.py
+++ b/python_on_whales/components/buildx/cli_wrapper.py
@@ -184,7 +184,7 @@ class BuildxCLI(DockerCLICaller):
         build_args: Dict[str, str] = {},
         builder: Optional[ValidBuilder] = None,
         cache: bool = True,
-        cache_from: Union[str, Dict[str, str], None] = None,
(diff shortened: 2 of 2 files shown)


tests/python_on_whales/components/buildx/test_buildx_cli_wrapper.py
```diff
@@ -102,6 +102,28 @@ def test_buildx_caching_dict_form(tmp_path):
...
+@pytest.mark.usefixtures("with_docker_driver")
+def test_buildx_caching_list_form(tmp_path):
+    (tmp_path / "Dockerfile").write_text(dockerfile_content1)
+    with docker.buildx.create(driver_options=dict(network="host"), use=True):
+        docker.buildx.build(
+            tmp_path,
+            cache_to=list(
+                dict(type="local", dest=tmp_path / "cache", mode="max"),
+                dict(type="local", dest=tmp_path / "cache2", mode="max"),
+            ),
```
gabrieldemarmiesse, line 114:
```suggestion
            cache_to=[
                dict(type="local", dest=tmp_path / "cache", mode="max"),
                dict(type="local", dest=tmp_path / "cache2", mode="max"),
            ],
```

tests/python_on_whales/components/buildx/test_buildx_cli_wrapper.py
```diff
@@ -102,6 +102,28 @@ def test_buildx_caching_dict_form(tmp_path):
...
+        )
+
+    with docker.buildx.create(driver_options=dict(network="host"), use=True):
+        docker.buildx.build(
+            tmp_path,
+            cache_from=list(
+                dict(type="local", src=tmp_path / "cache"),
+                dict(type="local", src=tmp_path / "cache2"),
+            ),
```
gabrieldemarmiesse, line 123:
```suggestion
            cache_from=[
                dict(type="local", src=tmp_path / "cache"),
                dict(type="local", src=tmp_path / "cache2"),
            ],
```

gabrieldemarmiesse (commented):
Thank for the pull request! I wasn't aware that this was possible. 

If this feature is only present in recent versions of buildx, you might have to update buildx in the CI: https://github.com/gabrieldemarmiesse/python-on-whales/blob/master/tests/Dockerfile#L3

tests/python_on_whales/components/buildx/test_buildx_cli_wrapper.py
redref: Done !

tests/python_on_whales/components/buildx/test_buildx_cli_wrapper.py
redref: Done !

redref:
Just checked and it seems the feature is in buildx since the beginning (at least 0.2.0 : https://github.com/docker/buildx/blob/v0.2.0/bake/bake.go#L134 ).

It is quite the only possibility to use buildx in a distributed caching behavior (from registry), as cache is overwritten and not appended if you push the same tag. On my side, I plan to use commit ids as cache tag, cache_to would be "<registry>/cache/<image_name>:<commit_id>", then cache_from would check each commit until reaching the base branch.

redref:
Oh it appears, only cache_from is supported, updated.
Finally figured how to run the tests locally, we should be fine now.

gabrieldemarmiesse:
Great, thanks a lot for the PR! 🎉  I'll make a new release this afternoon!

PR #238, opened 2021-08-25
Added a quiet flag to suppress the output of _pull_if_necessary.

gabriel-turturica:
I could not find a way to suppress the output of container.run when the image was not found.


gabrieldemarmiesse:
That's a good point. I'll think about it. Adding a functionnality that is not in the CLI must be done carefully. They have a very good user experience, and python-on-whales aims to ride on this very good UX

Issue #248, opened 2021-09-13
compose run


mdantonio:
Hello,
I know that a number of compose commands are still not implemented and I can imagine that you have your own priority list about them, but in case of tie can I ask you to push the run command higher in the ranking? :smile: 

I think that compose run is the last command that I need to complete the migration of all my use cases to python on whales, so the implementation of that command would be very appealing to me, but as usual I don't want to bother you with many requests, also considering that I have a number of other issues still open... so please consider this as a low priority request

Thank you again!

gabrieldemarmiesse:
What kind of options do you usually use with this command? That would help me put a priority on the options to implement the most useful first

mdantonio:
```
[COMMAND] [ARGS...]
--detach
--publish
--name
--no-deps
--rm
--user

```

gabrieldemarmiesse:
Thank you, I'll focus on those first!

mdantonio:
Thank to you!

mdantonio:
As reported in the PR comments the flag on compose run is --rm,
so `full_cmd.add_flag("--remove", remove)` should replaced with `full_cmd.add_flag("--rm", remove)`

The flag is therefore currently broken

```
The docker command executed was `/usr/local/bin/docker compose ....... --remove ....`.
It returned with code 16
The content of stdout is ''
The content of stderr is 'unknown flag: --remove
```

gabrieldemarmiesse:
I was sure I pushed the commit. It seems not. I'll fix it in the next PR

gabrieldemarmiesse:
thank you for reminding me. I'll make a release right now.

mdantonio:
Opened the PR #280 to add the stream option.
The tty flag was not working properly because it was not passed to the run wrapper, so changed:
```
run(full_cmd)
```

with

```
run(full_cmd, tty=tty)
```

mdantonio:
Hello,
I confirm that v0.31 fixes all my use cases based on compose run

Thank you for your support!

gabrieldemarmiesse:
I'm glad it works for you, thanks again for the pull request that's super helpful!

PR #250, opened 2021-09-23
Added docker.execute(interactive=True, tty=True)

gabrieldemarmiesse:
closes #194

PR #251, opened 2021-09-29
Added docker.compose.run()

gabrieldemarmiesse:
Initial implementation of docker.compose.run().

@mdantonio if you have some time to try it and give me feedback it would be great. I'm especially interested to know if it covers your use case.

Related to #248

### DIFF
--- a/python_on_whales/components/compose/cli_wrapper.py
+++ b/python_on_whales/components/compose/cli_wrapper.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import json
 from datetime import timedelta
+from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
 
 import python_on_whales.components.container.cli_wrapper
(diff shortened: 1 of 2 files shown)


python_on_whales/components/compose/cli_wrapper.py
```diff
@@ -215,9 +216,56 @@ def rm(
...
+        for port_mapping in publish:
+            if len(port_mapping) == 2:
+                full_cmd += ["--publish", f"{port_mapping[0]}:{port_mapping[1]}"]
+            else:
+                full_cmd += [
+                    "--publish",
+                    f"{port_mapping[0]}:{port_mapping[1]}/{port_mapping[2]}",
+                ]
+
+        full_cmd.add_flag("--remove", remove)
```
mdantonio, line 255: The flag on compose run is `--rm`, so this line should be:

```
full_cmd.add_flag("--rm", remove)
```

mdantonio:
Hello, of course I'm glad to test it! :smiley: 

Except for the --remove flag (commented above) the only issue that I can find is with long-lasting commands and interactive commands

For simple-noninteractve commands everything works well

```
>>> print(docker.compose.run(service="my_container", command=["ls"]))
a
b
c
```

For long-lasting commands the lack of a stream brings to a bad user experience. Let's image to the execute a script like:

test.sh
```
for i in $(seq 1 10); 
do 
	echo $i; 
	sleep 1
done
```

The expected output would be

1
... 1 second
2
... 1 second
3
... 1 second

and so on

but without the stream the result will be

... 10 seconds...
1
2
3
...


Interactive commands are not supported at all, but probably the some solution applied in #250 may work here too

I will be able to do some more extensive tests only after the release, so if you think that streams and interactive would require some long effort, then probably they could be included in a further improvement

Except for that (and the easily to be fixed remove flag), everything works as expected :+1:

gabrieldemarmiesse:
Thanks you for the feedback. It's indeed a good point that we should have a stream and an interactive option like we have on exec or run. I'll merge this PR, make a release and add those options later. I'll do it this weekend most likely.

PR #278, opened 2021-11-05
Fixed compose.restart by adding services and made timeout parameter optional

### LINKED ISSUE #277: docker compose restart
Hello,
another command that I would like to use (yes I know... my list of commands is pretty long :stuck_out_tongue: )
In the docs it is reported as _Not yet implemented_ but I found it mostly completed

I think that the only missing things are:
- the services parameter
- timeout should be defaulted to None

Here the corresponding commit on my fork

https://github.com/mdantonio/python-on-whales/commit/7bc1168dd8b2aaca51a67d5f6d1b677bc8e52d48

Can I ask you to consider the fixes? Also let me know if you prefer a PR and if I have to think about some tests

Thank you, as usual!

mdantonio:
Fixes #277

### DIFF
--- a/python_on_whales/components/compose/cli_wrapper.py
+++ b/python_on_whales/components/compose/cli_wrapper.py
@@ -183,7 +183,11 @@ class ComposeCLI(DockerCLICaller):
         full_cmd += services
         run(full_cmd)
 
-    def restart(self, services: Union[str, List[str]] = [], timeout: Union[int, timedelta, None] = None):
+    def restart(
+        self,
+        services: Union[str, List[str]] = [],
+        timeout: Union[int, timedelta, None] = None,
+    ):
         """Restart containers
 
         # Arguments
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
It seems there are issues with the CI. Looks like the code is not formatted with black. 

Can you run 
```
black ./
```
and push again? Thanks!

gabrieldemarmiesse (approved):

mdantonio:
Done.
Maybe that you want to consider to introduce a precommit configuration to enforce the use of black or other tools (isort, flake8, mypy)
(if you are interested I can open an issue and provide you with an example of configuration)

gabrieldemarmiesse:
I don't use pre-commit hooks but if you're willing to make a PR to add the pre-commit configuration of isort, flake8  and black, I'll be happy to merge it :)

I don't use/enforce mypy because the number of false negative is too great right now.

PR #280, opened 2021-11-06
Added stream option to docker.compose.run and fixed the tty flag

### LINKED ISSUE #248: compose run

mdantonio:
Fixes #248

### DIFF
--- a/python_on_whales/components/compose/cli_wrapper.py
+++ b/python_on_whales/components/compose/cli_wrapper.py
@@ -8,7 +8,7 @@ from typing import Any, Dict, List, Optional, Union
 import python_on_whales.components.container.cli_wrapper
 from python_on_whales.client_config import DockerCLICaller
 from python_on_whales.components.compose.models import ComposeConfig
-from python_on_whales.utils import run, to_list
+from python_on_whales.utils import run, to_list, stream_stdout_and_stderr
 
 
 class ComposeCLI(DockerCLICaller):
@@ -241,6 +241,7 @@ class ComposeCLI(DockerCLICaller):
         # labels: Dict[str, str] = {},
         name: Optional[str] = None,
         tty: bool = True,
+        stream: bool = False,
         dependencies: bool = True,
         publish: List[
             python_on_whales.components.container.cli_wrapper.ValidPortMapping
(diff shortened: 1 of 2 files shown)


mdantonio:
Due to the change from `run(full_cmd)` to `run(full_cmd, tty=tty)` the following test started to fail:

```
def test_compose_run_detach():
    container = docker.compose.run("alpine", ["echo", "dodo"], detach=True)

    time.sleep(0.1)
    assert not container.state.running
    assert container.logs() == "dodo\n"
```

This is because the run wrapper with tty enabled directs the output to sys.stdout/err instead of subprocess.PIPE, so that the test is unable to obtain the container name

Disabled the tty ( => `container = docker.compose.run("alpine", ["echo", "dodo"], detach=True, tty=False)`) to fix the test, I hope that it is the correct approach

python_on_whales/components/compose/cli_wrapper.py
```diff
@@ -252,7 +253,41 @@ def run(
...
+            raise ValueError(
+                "You can't set tty=True and stream=True at the same"
+                "time. Their purpose are not compatible."
+            )
+
+        if detach and stream:
+            raise ValueError(
+                "You can't detach and stream at the same time. It's not compatible."
+            )
+
```
gabrieldemarmiesse, line 290: I would also add here this:

```suggestion
            )

    if detach and tty:
            raise ValueError(
                "You can't detach and set tty=True at the same time. It's not compatible. "
                "Try setting tty=False in docker.compose.run(...)."
            )
```

python_on_whales/components/compose/cli_wrapper.py
```diff
@@ -252,7 +253,41 @@ def run(
...
+        # Returns:
+            Optional[str]
+
+        """
+
+        if tty and stream:
+            raise ValueError(
+                "You can't set tty=True and stream=True at the same"
+                "time. Their purpose are not compatible."
```
gabrieldemarmiesse, line 283:
```suggestion
                "time. Their purpose are not compatible. Try setting tty=False in docker.compose.run"
```

gabrieldemarmiesse (changes requested):
A few changes to be made, but it's very nice overall :) thanks for the pull request!

gabrieldemarmiesse (approved):

PR #284, opened 2021-11-23
Enable labeling networks in network create

Akshay04:
Add missing functionality to docker network create function to add labels to network

### DIFF
--- a/python_on_whales/components/network/cli_wrapper.py
+++ b/python_on_whales/components/network/cli_wrapper.py
@@ -155,6 +155,7 @@ class NetworkCLI(DockerCLICaller):
         driver: Optional[str] = None,
         gateway: Optional[str] = None,
         subnet: Optional[str] = None,
+        labels: Dict[str, str] = {},
         options: List[str] = [],
     ) -> Network:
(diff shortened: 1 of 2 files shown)


tot0:
Awesome thanks!

gabrieldemarmiesse:
That's great! Thanks a lot! Could you add a small unit test? :)

Akshay04:
Added the unit test, I was having some trouble testing my changes locally. How do you typically do this?

gabrieldemarmiesse:
You need to install the linters and formatters:
Isort, black and flake 8. Run the at the root of the repo to check things.

Next to run the tests, install this package in editable mode `pip install -e . ` from the root of the repo too.

Install the test requirements too
`pip install -r tests/requirements.txt`

Then run a test with
`pytest path/to/file.py::function_to_run`.

The test looks good. Could you just create the network with a context manager so that it's deleted afterwards even if there are exceptions?
`with docker.network.create(...) as my_network:`

Akshay04:
Will do, thanks for your help

gabrieldemarmiesse (approved):

PR #298, opened 2022-01-13
Add compose_profiles argument to the docker client constructor

gabrieldemarmiesse:


parity3:
I'm just going to link this here for users of docker compose 2 and possibly other implementations with profiles support, that docker compose down will ignore the profiles passed in and remove all services (but the rest of the commands DO observe profiles):

https://github.com/docker/compose/issues/8139

It may also be advisable to add a warning in this library's docs but that may be going too far. I'll leave that to you @gabrieldemarmiesse !

Thanks again!

gabrieldemarmiesse:
Thanks for the message! You can always make a pull request to change the documentation of python-on-whales and add a note there. 

I don't think it should be up to me to add a warning. I would prefer to simply make a CLI wrapper and nothing more (expecially not take too many UX decisions). If python-on-whales were to display warnings, it would be when the behavior diverges from the CLI one. It's up to docker compose to add warnings :)

parity3:
To clarify, I wasn't talking about a code-level warning or user message displayed during execution. I was just talking about a blurb somewhere in the README or even tucked away in the examples page of the documentation. But we may already be on that same page.  Re: this is a docker cli issue, I totally agree with your points. And it'd be beyond scope for this project to try and follow and advise on CLI bugs.

It's just a non-obvious painpoint I've had to work around, but may still not really apply to a general audience. This issue is probably enough!

Thanks again for this project! I am using it heavily as a tool in my ci/cd chain.

gabrieldemarmiesse:
I'm glad you like it! It may deserve a not in the doc as you're saying.

PR #305, opened 2022-02-03
Support for podman (dropin docker replacement)

joshbode:
- podman doesn't support `buildx inspect` so provide default values
- change `image remove` to `image rm` (note: this is the documented
  command for both `docker` and `podman` - see
  https://docs.docker.com/engine/reference/commandline/image_rm/)


joshbode:
Related issue: [podman/issues/13014](https://github.com/containers/podman/issues/13014)

gabrieldemarmiesse:
I am ok with such a feature but we need some way of testing it in the CI. Ideally all tests should run twice, once with docker and once with podman. It's fine if it's only a subset of all tests at first.

Note that docker will be a first class citizen and as such, we won't change any of the class names and methods.

joshbode:
Definitely! I'll work on that this week and see if I can get something working in the tests - I suspect there are more examples of missing commands in `podman`, although the upcoming release appears to be more complete (especially for buildx)

python_on_whales/components/buildx/cli_wrapper.py
```diff
@@ -81,6 +84,18 @@ def remove(self):
         """
         BuildxCLI(self.client_config).remove(self)
 
+    @property
+    def implementation(self) -> str:
```
gabrieldemarmiesse, line 88: I believe this information should go in the https://github.com/gabrieldemarmiesse/python-on-whales/blob/master/python_on_whales/client_config.py#L45

The api should look like this
```python
from python_on_whales import DockerClient
podman = DockerClient(client_binary="podman")
output = podman.run("hello-world")
print(output)
```

In this way, that would also work with nerdctl for example.

You can also modify https://github.com/gabrieldemarmiesse/python-on-whales/blob/master/python_on_whales/client_config.py#L64 so that it will search for the "client_binary" instead of docker

joshbode:
Time got away from me on this PR, but I'm going to work on the tests next :)

gabrieldemarmiesse:
I have made a PR for this, I didn't make an official release yet but you can see the usage: https://github.com/gabrieldemarmiesse/python-on-whales#alternatives-to-docker-podman-nerdctl

gabrieldemarmiesse:
I'll close this PR as the functionnality seems to be available in master. Feel free to reopen it if needed!

PR #309, opened 2022-02-26
Add project_urls

wannaphong:
Add project_urls from https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls

The string of the key is the exact text that will be displayed on PyPI.


gabrieldemarmiesse (approved):
Thanks a lot!

PR #314, opened 2022-03-18
Add flags to docker compose pull

joshkarlin:
This PR adds the following flags to `docker.compose.pull` as optional keyword arguments:
```
--ignore-pull-failures
--include-deps
```

The test for `docker.compose.pull` skips so I have added the same skip to the new tests.  Presumably they would fail for the same reason

### DIFF
--- a/python_on_whales/components/compose/cli_wrapper.py
+++ b/python_on_whales/components/compose/cli_wrapper.py
@@ -245,15 +245,25 @@ class ComposeCLI(DockerCLICaller):
         Container = python_on_whales.components.container.cli_wrapper.Container
         return [Container(self.client_config, x, is_immutable_id=True) for x in ids]
 
-    def pull(self, services: List[str] = []):
+    def pull(
+        self,
+        services: List[str] = [],
+        ignore_pull_failures: bool = False,
+        include_deps: bool = False,
+    ):
         """Pull service images
 
         # Arguments
             services: The list of services to select. Only the images of those
                 services will be pulled. If no services are specified (the default
                 behavior) all images of all services are pulled.
+            ignore_pull_failures: Pull what it can and ignores images with pull failures
+            include_deps: Also pull services declared as dependencies
+
         """
         full_cmd = self.docker_compose_cmd + ["pull"]
+        full_cmd.add_flag("--ignore-pull-failures", ignore_pull_failures)
+        full_cmd.add_flag("--include-deps", include_deps)
         full_cmd += services
         run(full_cmd)
 
--- a/tests/python_on_whales/components/dummy_compose.yml
+++ b/tests/python_on_whales/components/dummy_compose.yml
@@ -17,11 +17,9 @@ services:
     command: sleep infinity
   busybox-2-electric-boogaloo:
     image: busybox:latest
-    depends:
+    depends_on:
       - alpine
     command: sleep infinity
-  ghost:
-    image: i_do_not_exist:v0.0.1
   alpine:
     image: alpine:latest
     command: sleep infinity
--- a/tests/python_on_whales/components/dummy_compose.yml
+++ b/tests/python_on_whales/components/dummy_compose.yml
@@ -15,6 +15,13 @@ services:
   busybox:
     image: busybox:latest
     command: sleep infinity
+  busybox-2-electric-boogaloo:
+    image: busybox:latest
+    depends:
+      - alpine
+    command: sleep infinity
+  ghost:
+    image: i_do_not_exist:v0.0.1
   alpine:
     image: alpine:latest
     command: sleep infinity
--- a/tests/python_on_whales/components/dummy_compose_non_existent_image.yml
+++ b/tests/python_on_whales/components/dummy_compose_non_existent_image.yml
@@ -1,0 +1,8 @@
+version: "3.7"
+
+services:
+  busybox:
+    image: busybox:latest
+    command: sleep infinity
+  ghost:
+      image: busybox:i-do-not-exist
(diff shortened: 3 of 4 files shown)


gabrieldemarmiesse:
Do you need help? It would be cool to have those flags available :)

joshkarlin:
i'm having a bit of trouble with the tests.  some of the tests i haven't touched fail, for example:
```
>       assert container_names == {"dudu_busybox_1", "dudu_alpine_1"}
E       AssertionError: assert {'dudu-alpine...du-busybox-1'} == {'dudu_alpine...du_busybox_1'}
E         Extra items in the left set:
E         'dudu-busybox-1'
E         'dudu-alpine-1'
E         Extra items in the right set:
E         'dudu_busybox_1'
E         'dudu_alpine_1'
E         Use -v to get more diff
```

gabrieldemarmiesse:
You are right. Docker compose recently switch from _ to a -. I believe there is a flag to handle this and revert it. I'll fix the tests. Thank you for reporting it :)

joshkarlin:
no worries!  i guess the change is in how they automatically generate `container_name` if it isn't specified?

i like this library compared to `docker-py` so you may see more contributions from me to fill in last remaining gaps

gabrieldemarmiesse:
I'm always happy to get new flags and features :)

joshkarlin:
would you like me to fix them?  can change _ to - but then tests may end up passing or failing depending on docker version installed...

joshkarlin:
looks like compose v2 uses hyphens, maybe best to change from underscore to hyphen?

gabrieldemarmiesse:
I know there is a docker compose flag to choose if we want to use _ or -. We should add this flag to oython on whales and use this flag in our tests. With this the unit tests will pass, independently if the docker compose version ( I think). I'll implement it eventually when I have the time. Otherwise if you're motivated you can do it too. You may have to upgrade the dockercompose used in the CI. You can choose the version when the plugin is downloaded with wget (it happens in a Dockerfile).

joshkarlin:
the flag is `--compatibility` which suggests more changes than just `-` -> `_`, unsure on everything it does as it doesn't seem particularly well documented:  https://docs.docker.com/compose/cli-command-compatibility/#flags-that-will-not-be-implemented

people have asked for more granular control but that hasn't been added: https://github.com/docker/compose/issues/9061

I will implement this in the place where `docker compose` commands are constructed and submit a PR soon

gabrieldemarmiesse:
If you think you can make the CI green, feel free to reopen this PR yes :)

joshkarlin:
Great.  The CI should be green if it was before this PR.  The tests I added are skipped as this is what is happening for the existing `docker.compose.pull` test.

I have installed locally and tested the added args.

`include_deps`:
```python
>>> docker = DockerClient(compose_files=["dummy_compose.yml"])
>>> docker.image.exists("alpine")
False
>>> docker.image.exists("busybox")
False
>>> docker.compose.pull(["busybox-2-electric-boogaloo"])
>>> docker.image.exists("busybox")
True
>>> docker.image.exists("alpine")
False
>>> docker.compose.pull(["busybox-2-electric-boogaloo"], include_deps=True)
>>> docker.image.exists("alpine")
True
```
And `ignore_pull_failures`:
```python
>>> docker = DockerClient(compose_files=["dummy_compose_non_existent_image.yml"])
>>> docker.compose.pull(["ghost"])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/josh.karlin/codes/python-on-whales/python_on_whales/components/compose/cli_wrapper.py", line 268, in pull
    run(full_cmd)
  File "/home/josh.karlin/codes/python-on-whales/python_on_whales/utils.py", line 150, in run
    raise DockerException(
python_on_whales.exceptions.DockerException: The docker command executed was `/usr/bin/docker compose --file dummy_compose_non_existent_image.yml pull ghost`.
It returned with code 18
The content of stdout is ''
The content of stderr is 'ghost Pulling 
ghost Error 
WARNING: Some service image(s) must be built from source by running:
    docker compose build %s ghost
Error response from daemon: pull access denied for i-definitely-do-not-exist, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
'

>>> docker.compose.pull(["ghost"], ignore_pull_failures=True)
>>> 
```
Obviously would prefer the tests to actually run.  Any idea on the fix for the issue which requires skipping the tests?  Does it just need the `HOME` var to be defined?

gabrieldemarmiesse:
Weird, usually HOME is defined on nearly all linux environments. If you run `echo $HOME` what do you get?

gabrieldemarmiesse:
So I believe that it should work in the CI, I had a similar issue a while ago and I fixed it by adding this line in the Dockerfile: https://github.com/gabrieldemarmiesse/python-on-whales/blob/master/tests/Dockerfile#L38

Maybe you had this issue locally but it won't happen in the CI?

joshkarlin:
I didn't actually have the issue locally, I was going by the comment in the existing test but if that's not an issue then great!  I will remove the `skipif`.  Looks like your fix did the trick :+1:

gabrieldemarmiesse (approved):

gabrieldemarmiesse:
I hope it will work :) Thanks for the PR, great addition!

gabrieldemarmiesse:
@joshkarlin do you want to make other PRs before I make a release? Otherwise you can always do PRs later on.

joshkarlin:
Thanks for asking!  There is one thing I noticed just now which raised a question, what is the logic for choosing whether or not methods for a `DockerCLICaller` subclass call `run` with `capture_stdout` and `capture_stderr` as `True` vs `False`?

gabrieldemarmiesse:
It's depending on if we use the output or not. We don't want to display any output unless it's for a TTY. This has to be done on a per-command basis.

PR #338, opened 2022-05-31
Handle custom registry name when using ecr login

a-gentilhomme:
This PR adds support for a custom registry name when using ECR login. 
It can be used like so : 

```python
    target_registry = "my-very-special-registry"
    docker.login_ecr(
        region_name="eu-west-1",
        registry=target_registry,
    )
    docker.pull(f"{target_registry}/my-special-project:my-special-tag")
```

### DIFF
--- a/python_on_whales/docker_client.py
+++ b/python_on_whales/docker_client.py
@@ -236,7 +236,7 @@ class DockerClient(DockerCLICaller):
         aws_access_key_id: Optional[str] = None,
         aws_secret_access_key: Optional[str] = None,
         region_name: Optional[str] = None,
-        registry: Optional[str] = None
+        registry: Optional[str] = None,
     ):
         """Login to the aws ECR registry. Credentials are taken from the
         environment variables as defined in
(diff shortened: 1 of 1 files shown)


python_on_whales/docker_client.py
```diff
@@ -236,6 +236,7 @@ def login_ecr(
         aws_access_key_id: Optional[str] = None,
         aws_secret_access_key: Optional[str] = None,
         region_name: Optional[str] = None,
+        registry: Optional[str] = None
```
gabrieldemarmiesse, line 239: Very nice, can you add some information in the docstring about this argument? It will be displayed later on in the docs at https://gabrieldemarmiesse.github.io/python-on-whales/docker_client/#login_ecr

gabrieldemarmiesse (changes requested):

python_on_whales/docker_client.py
```diff
@@ -250,6 +251,11 @@ def login_ecr(
         ```
 
         You need botocore to run this function. Use `pip install botocore` to install it.
+
+        The `registry` parameter can be used to override the registry that is guessed from the authorization token
+        request's response.
+        It is especially useful if the aws account you use can access several repositories and you
```
gabrieldemarmiesse, line 257:
```suggestion
        The `registry` parameter can be used to override the registry that is guessed from the authorization token
        request's response.
        In other words: If the registry is `None` (the default) then it will be assumed that it's the ECR registry linked to the credentials provided.
        It is especially useful if the aws account you use can access several repositories and you
```

gabrieldemarmiesse (approved):

PR #340, opened 2022-06-23
[FEAT] Add Event API

lonk95:
Added a simple implementation for the events api

### DIFF
--- a/python_on_whales/components/system/cli_wrapper.py
+++ b/python_on_whales/components/system/cli_wrapper.py
@@ -96,13 +96,9 @@ class SystemCLI(DockerCLICaller):
         full_cmd.add_simple_arg("--until", format_time_arg(until))
         full_cmd.add_args_list("--filter", format_dict_for_cli(filters))
         iterator = stream_stdout_and_stderr(full_cmd)
-        event_function = (
-            lambda stream_tuple: DockerEvent.parse_raw(stream_tuple[1])
-            if stream_tuple[0] == "stdout"
-            else None
-        )
-        event_iterator = map(event_function, iterator)
-        return event_iterator
+        for stream_origin, stream_content in iterator:
+            if stream_origin == "stdout":
+                yield DockerEvent.parse_raw(stream_content)
 
     def info(self) -> SystemInfo:
         """Returns diverse information about the Docker client and daemon.
--- a/python_on_whales/components/system/cli_wrapper.py
+++ b/python_on_whales/components/system/cli_wrapper.py
@@ -61,7 +61,7 @@ class SystemCLI(DockerCLICaller):
         full_cmd = self.docker_cmd + ["system", "df", "--format", "{{json .}}"]
         return DiskFreeResult(run(full_cmd))
 
-    def events(self, filters:List[str]) -> List[DockerEvent]:
+    def events(self, filters: List[str]) -> List[DockerEvent]:
         """Returns docker events information up to the current point in time.
 
         # Arguments
@@ -78,7 +78,14 @@ class SystemCLI(DockerCLICaller):
         [reference page for
         system events](https://docs.docker.com/engine/api/v1.40/#operation/SystemEvents)
         """
-        full_cmd = self.docker_cmd + ["system", "events", "--format", "{{json .}}", "--until", "0s"]
+        full_cmd = self.docker_cmd + [
+            "system",
+            "events",
+            "--format",
+            "{{json .}}",
+            "--until",
+            "0s",
+        ]
         for filter in filters:
             full_cmd.extend(["--filter", filter])
         event_strings = run(full_cmd).split()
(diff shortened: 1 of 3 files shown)


gabrieldemarmiesse:
thanks a lot for the pull request! I'll take a look :)

python_on_whales/components/system/cli_wrapper.py
```diff
@@ -57,9 +61,39 @@ def disk_free(self) -> DiskFreeResult:
         full_cmd = self.docker_cmd + ["system", "df", "--format", "{{json .}}"]
         return DiskFreeResult(run(full_cmd))
 
-    def events(self):
-        """Not yet implemented"""
-        raise NotImplementedError
+    def events(self, filters: List[str]) -> List[DockerEvent]:
```
gabrieldemarmiesse, line 64:
```suggestion
    def events(self, until: Union[None, datetime.datetime, datetime.timedelta], filters: Dict[str, str]) -> Iterator[DockerEvent]:
```
The main usage of `docker system events` is streaming, as such Iterator will only work in this case. In the case where we have `--until=...` where the listening loop will exit at some point, an iterator still works. 

If the user wants a list, the user should do `list(docker.system.events(...)`.

I believe it might be a bit too hard/confusing for users to only support the case where the function will end at some point. As such, I would prefer that we implement directly the general case (iterator of infinite event) and the user will be able to pass the `until` argument if he/she wants the iterator to stop at some point.

I will guide you to implement this if you're willing to work on it :)

For the filters, I would like to use a dict to be consitent with the other functions. You can take as reference the implementation of `docker.container.prune(...)`.

For the "until" argument, we have something similar in `docker.container.logs()`, we use a datetime or a timedelta. Please do the same here. You can copy the code as needed.

We should return an iterator, as such we need to use the same code as in functions like `docker.run(stream=True)` or `docker.container.logs()`. You should use the function ` stream_stdout_and_stderr` instead of `run`.

As you get the stdout out of `stream_stdout_and_stderr`, you will be able to convert it into a `DockerEvent` and then `yield` it to the user.

tests/python_on_whales/components/test_system.py
```diff
@@ -23,6 +24,20 @@ def test_info():
     assert "local" in info.plugins.volume
 
 
+def test_events():
```
gabrieldemarmiesse, line 27: Very cool to have added a unit test!

tests/python_on_whales/components/test_system.py
```diff
@@ -23,6 +24,20 @@ def test_info():
...
 
+def test_events():
+    name = random_name()
+    docker.run("hello-world", remove=True, name=name)
+    # Takes some time for events to register
+    sleep(1)
+    events = docker.system.events(filters=[f"container={name}"])
+    # Check that we capture all the events from container create to destroy
+    assert len(events) >= 5
+    actions = []
```
gabrieldemarmiesse, line 35:
```suggestion
    actions = set()
```
actions can be a set, there is no need to use a list first and then convert it to a set

gabrieldemarmiesse (changes requested):
Thanks again for this pull request, this is a very cool functionnality! We'll need to rework the implementation to return an iterator, so it's more work, but it will allow us to handle many more use cases easily :) 

Feel free to ask if something is hard to do or not clear enough

python_on_whales/components/system/cli_wrapper.py
lonk95: Thanks for the feedback! I'll work on a revision.

tests/python_on_whales/components/test_system.py
lonk95: Done

tests/python_on_whales/components/test_system.py
lonk95: Thanks!

python_on_whales/components/system/cli_wrapper.py
lonk95: Pushed some changes to match what you suggested.

python_on_whales/components/system/cli_wrapper.py
```diff
@@ -57,9 +67,42 @@ def disk_free(self) -> DiskFreeResult:
...
+            "system",
+            "events",
+            "--format",
+            "{{json .}}",
+        ]
+        full_cmd.add_simple_arg("--since", format_time_arg(since))
+        full_cmd.add_simple_arg("--until", format_time_arg(until))
+        full_cmd.add_args_list("--filter", format_dict_for_cli(filters))
+        iterator = stream_stdout_and_stderr(full_cmd)
+        event_function = (
```
gabrieldemarmiesse, line 99: I believe using the `yield` keyword would make the code easier to read. You can do something like this:
```python
for stream_origin, stream_content in stream_stdout_and_stderr(...):
    if stream_origin == "stdout":
        yield DockerEvent.parse_raw(stream_content)
```

gabrieldemarmiesse (changes requested):
Great PR! I think we just need to clarify the code and we're good to go!

python_on_whales/components/system/cli_wrapper.py
lonk95: done

gabrieldemarmiesse (approved):
Thanks!

PR #342, opened 2022-07-07
Allow to inspect exit_code and other info on exceptions

juan-leon:
DockerException classes format the relevant info nicely for user.

But sometimes code needs to use that info to take decisions or to digest that
for user.

For example, exit code of some `docker compose exec ...` command might be
relevant, and we would want to implement some logic that depends on exit code or
parse a json in stderr, etc.

### DIFF
--- a/docs/template/user_guide/exceptions.md
+++ b/docs/template/user_guide/exceptions.md
@@ -1,0 +1,35 @@
+# Capturing exceptions
+
+## Exception classes
+
+Exceptions raised will be an instance of `DockerException`, or a child class for
+more specific errors.
+
+Those are the child classes:
+
+* `NoSuchContainer`
+* `NoSuchImage`
+* `NoSuchService`
+* `NotASwarmManager`
+* `NoSuchVolume`
+
+All exceptions will have these 4 attributes:
+
+* docker_command: the docker command used internally, as a list of strings.
+* return_code: the exit code docker client exited with, as an int
+* stdout: the content that docker wrote to stdout, as a string, or `None`
+* stderr: the content that docker wrote to stderr, as a string, or `None`
+
+## Example
+
+```python
+import logging
+from python_on_whales import DockerClient
+from python_on_whales.exceptions import DockerException
+
+client = DockerClient(compose_files=["/tmp/docker-compose.yml"])
+try:
+    client.execute("my-service", ["arg1", "arg2"])
+except DockerException as e:
+    print(f"Exit code {e.return_code} while running {e.docker_command}")
+```
--- a/python_on_whales/exceptions.py
+++ b/python_on_whales/exceptions.py
@@ -9,20 +9,31 @@ class DockerException(Exception):
         stdout: Optional[bytes] = None,
         stderr: Optional[bytes] = None,
     ):
+
+        self.docker_command: List[str] = command_launched
+        self.return_code: int = return_code
+        if stdout is None:
+            self.stdout: Optional[str] = None
+        else:
+            self.stdout: Optional[str] = stdout.decode()
+        if stderr is None:
+            self.stderr: Optional[str] = None
+        else:
+            self.stderr: Optional[str] = stderr.decode()
         command_launched_str = " ".join(command_launched)
         error_msg = (
             f"The docker command executed was `{command_launched_str}`.\n"
             f"It returned with code {return_code}\n"
         )
         if stdout is not None:
-            error_msg += f"The content of stdout is '{stdout.decode()}'\n"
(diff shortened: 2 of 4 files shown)


python_on_whales/exceptions.py
```diff
@@ -9,20 +9,23 @@ def __init__(
         stdout: Optional[bytes] = None,
         stderr: Optional[bytes] = None,
     ):
-        command_launched_str = " ".join(command_launched)
+        self.docker_command = " ".join(command_launched)
+        self.return_code = return_code
+        self.stderr = stderr.decode() if stderr else ""
+        self.stdout = stdout.decode() if stdout else ""
```
gabrieldemarmiesse, line 15: A few changes would be needed to improve the user experience :) 

1) All attributes should be typed, it will help users when coding in their IDEs or when using Mypy. Just follow the syntax above
2) We should use more explicit types, things that the user can interact with easily and programmatically. For exemple:
`self.docker_command: List[str] = command_launched` because if you join the list, you will loose valuable information, for exemple, where the arguments start or end if there is a space in them.
3) self.stderr and self.stdout should be None and not "". It will allow the user to differentiate between an stderr not captured, and an empty stderr
```python
if stderr is None:
    self.stderr: Optional[str] = None
else:
    self.stderr: Optional[str] = stderr.decode()
```

python_on_whales/exceptions.py
```diff
@@ -9,20 +9,23 @@ def __init__(
...
         error_msg = (
-            f"The docker command executed was `{command_launched_str}`.\n"
+            f"The docker command executed was `{self.docker_command}`.\n"
```
gabrieldemarmiesse, line 17: Please add a small unit test where you inspect the exception and the attributes

python_on_whales/exceptions.py
```diff
@@ -9,20 +9,23 @@ def __init__(
...
-            error_msg += f"The content of stdout is '{stdout.decode()}'\n"
+            error_msg += f"The content of stdout is '{self.stdout}'\n"
         else:
             error_msg += (
                 "The content of stdout can be found above the "
                 "stacktrace (it wasn't captured).\n"
             )
         if stderr is not None:
-            error_msg += f"The content of stderr is '{stderr.decode()}'\n"
+            error_msg += f"The content of stderr is '{self.stderr}'\n"
```
gabrieldemarmiesse, line 28: Can you provide a small user guide explaining the attributes available? 
Just add a markdown file here: https://github.com/gabrieldemarmiesse/python-on-whales/tree/master/docs/template/user_guide

and declare it here: https://github.com/gabrieldemarmiesse/python-on-whales/blob/master/docs/mkdocs.yml

thanks!

gabrieldemarmiesse (changes requested):
Thank you for the pull request! I think it's a great addition :)

python_on_whales/exceptions.py
juan-leon: Thanks for the suggestion.  Done in last version or PR (c810682).

python_on_whales/exceptions.py
juan-leon: Done in last version or PR (c810682).

python_on_whales/exceptions.py
juan-leon: Done in last version or PR (c810682).

docs/template/user_guide/exceptions.md
```diff
@@ -0,0 +1,35 @@
...
+```python
+import logging
+from python_on_whales import DockerClient
+from python_on_whales.exceptions import DockerException
+
+client = DockerClient(compose_files=["/tmp/docker-compose.yml"])
+try:
+    client.execute("my-service", ["arg1", "arg1"])
+except DockerException as e:
+    print("Exit code %s while running %s", e.return_code, e.docker_command)
```
gabrieldemarmiesse, line 34: Please use fstring formatting for readability

gabrieldemarmiesse (changes requested):
Extremely cool :) just a small change and we can merge!

docs/template/user_guide/exceptions.md
juan-leon: Done at 9f350b7

gabrieldemarmiesse (approved):
Awesome PR!

gabrieldemarmiesse:
@juan-leon I made a new release, you should be able to use the feature you implemented. 

The new doc is live :) https://gabrieldemarmiesse.github.io/python-on-whales/user_guide/exceptions/

juan-leon:
Thanks for the quick release and the improvement suggestions!

PR #343, opened 2022-07-10
Allow interactive container to be detached

### LINKED ISSUE #344: Allow interactive container to be detached
Running a container with `--interactive` (or `-i`) causes stdin to be open. This may be desirable for reasons entirely separate from whether you want to be attached/detached. For example, with `bash` as the entrypoint the container will exit if stdin is closed.

This is a hard blocker for my use-case, so I'm currently having to use a fork with this `ValueError` removed as per #343 (feel free to take over that draft PR, or let me know what's needed to get it ready for merge).

LewisGaul:
Running a container with `--interactive` (or `-i`) causes stdin to be open. This may be desirable for reasons entirely separate from whether you want to be attached/detached. For example, with `bash` as the entrypoint the container will exit if stdin is closed.

This is a hard blocker for my use-case, so I'm currently having to use a fork with this `ValueError` removed. It would be great if this could be upstreamed! Possibly even remove the `if stream and interactive` error case too (I'm less familiar with the streaming case).

Fixes #344

- [x] Add UT case

### DIFF
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -1444,12 +1444,6 @@ class ContainerCLI(DockerCLICaller):
 
         full_cmd.add_flag("--init", init)
 
-        # TODO: activate interactive and tty
-        if interactive and not tty:
-            raise NotImplementedError(
-                "Currently, docker.container.run(interactive=True) must have"
-                "tty=True. interactive=True and tty=False is not yet implemented."
-            )
         full_cmd.add_flag("--interactive", interactive)
         full_cmd.add_flag("--tty", tty)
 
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse:
Thank you for the pull request! I'm not familiar with this use case but if you use it, then it must work :)

Could you add in the unit tests a very simple example of how you would use it? With such a unit test, we would ensure your use case continues working even if we do big refactorings in the future

LewisGaul:
I've removed a couple more restrictions involving `interactive` - we actually normally run tests with `tty=False`, but `interactive=True` is required for our containers, and this combination was being disallowed as well.

gabrieldemarmiesse (approved):
Looks good! Thanks!

gabrieldemarmiesse:
Can we merge or do you want to add something else?

LewisGaul:
> Can we merge or do you want to add something else?

Feel free to merge :)

PR #346, opened 2022-07-10
Implement 'docker build' as alternative to 'docker buildx build'

### LINKED ISSUE #345: Use main docker build command by default rather than buildx
Currently `docker.build()` is an alias to `docker.buildx.build()`, but it's not clear to me why this is since `docker build` and `docker buildx build` aren't entirely compatible and the latter still seems to be somewhat experimental?

I have hit two issues with `docker.buildx.build()`: using `FROM sha256:{image.id}` in a Dockerfile is not recognised (it is with `docker build`); and on another host my docker client does not have buildx installed (and I'm very reluctant for this to be a requirement).

Under the assumption that the main reason `docker build` is not implemented is because it's additional effort on top of supporting `docker buildx` commands I've put together a basic implementation that is working for me - the PR is #346, feel free to take it over or let me know how I can get it ready for merge. Unfortunately this is a breaking change, since the `docker.build()` API is slightly different... This does seem a more correct mirror of the CLI, so I wonder if there's a way to get it in with the appropriate changelog notices and version bump?

LewisGaul:
Currently `docker.build()` is an alias to `docker.buildx.build()`, but it's not clear to me why this is since `docker build` and `docker buildx build` aren't entirely compatible and the latter still seems to be somewhat experimental?

I have hit two issues with `docker.buildx.build()`: using `FROM sha256:{image.id}` in a Dockerfile is not recognised (it is with `docker build`); and on another host my docker client does not have buildx installed (and I'm very reluctant for this to be a requirement).

Under the assumption that the main reason `docker build` is not implemented is because it's additional effort on top of supporting `docker buildx` commands I've put together a basic implementation that is working for me - feel free to take it over or let me know how I can get it ready for merge.

Planned to fix #345

- [x] Doc updates
- [x] UT updates


python_on_whales/components/image/cli_wrapper.py
```diff
@@ -202,6 +202,80 @@ def __init__(self, client_config: ClientConfig):
...
 
+    def legacy_build(
+        self,
+        context_path: ValidPath,
+        add_hosts: Dict[str, str] = {},
+        build_args: Dict[str, str] = {},
+        cache: bool = True,
+        file: Optional[ValidPath] = None,
+        labels: Dict[str, str] = {},
+        network: Optional[str] = None,
```
LewisGaul, line 213: I think maybe this should accept a list? Not sure

python_on_whales/components/image/cli_wrapper.py
```diff
@@ -202,6 +202,80 @@ def __init__(self, client_config: ClientConfig):
...
+        A `python_on_whales.Image` is returned, even when using multiple tags.
+        That is because it will produce a single image with multiple tags.
+
+        # Arguments
+            context_path: The path of the build context.
+            add_hosts: Hosts to add. `add_hosts={"my_host1": "192.168.32.35"}`
+            build_args: The build arguments.
+                ex `build_args={"PY_VERSION": "3.7.8", "UBUNTU_VERSION": "20.04"}`.
+            cache: Whether or not to use the cache, defaults to True
+            file: The path of the Dockerfile
```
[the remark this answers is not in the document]
gabrieldemarmiesse: We could add that indeed.

python_on_whales/components/image/cli_wrapper.py
gabrieldemarmiesse: The `docker image build --help` seems to disallow multiple inputs, so I believe multiple networks only works with buildx.

gabrieldemarmiesse:
@LewisGaul could you take one last look before I merge? To see if there is something I forgot in your use case?

PR #347, opened 2022-07-11
Fix versions for linting and pytest

gabrieldemarmiesse:


tests/test-requirements.txt
```diff
@@ -1,6 +1,6 @@
-pytest
-flake8
-black
-isort
+pytest==7.*
+flake8==4.0.1
+black==22.6.0
```
LewisGaul, line 3: https://github.com/gabrieldemarmiesse/python-on-whales/blob/master/.pre-commit-config.yaml#L7 needs an update too?

gabrieldemarmiesse: you are totally right!

PR #351, opened 2022-07-12
Improve legacy build parsing of image ID

LewisGaul:
Apologies for not realising this was a better way to do it from the start... It turns out the original method doesn't work with podman, which gives slightly different output when not using `--quiet`.

### DIFF
--- a/python_on_whales/components/image/cli_wrapper.py
+++ b/python_on_whales/components/image/cli_wrapper.py
@@ -252,7 +252,7 @@ class ImageCLI(DockerCLICaller):
         # to make it easier to write and read tests, the tests of this function
         # are also grouped with the tests of "docker.build()".
         tags = to_list(tags)
-        full_cmd = self.docker_cmd + ["build"]
+        full_cmd = self.docker_cmd + ["build", "--quiet"]
 
         full_cmd.add_args_list(
             "--add-host", format_dict_for_cli(add_hosts, separator=":")
@@ -270,10 +270,7 @@ class ImageCLI(DockerCLICaller):
             self.client_config
         )
         full_cmd.append(context_path)
-        stdout_lines = run(full_cmd).splitlines()
-        for line in stdout_lines:
-            if "Successfully built" in line:
-                image_id = line.split(" ")[-1]
+        image_id = run(full_cmd).strip()
         return docker_image.inspect(image_id)
 
     def history(self):


gabrieldemarmiesse (changes requested):
Thank you for the pull request!

Sadly here, if an issue occurs, we will loose some debugging information in the exception as the quiet flag is always used. For podman, is it possible to parse the image id without the quiet flag?

LewisGaul:
> Sadly here, if an issue occurs, we will loose some debugging information in the exception as the quiet flag is always used.

Good point that this is a concern, however in the case the build errors the relevant output will go to stderr and will still be included in the exception. Examples below.

Docker output:
```
root@ubuntu:~/rebuild# docker build . -q > /dev/null
Sending build context to Docker daemon  2.048kB
Step 1/5 : FROM ubuntu AS stage1
 ---> 7c73d141098a
Step 2/5 : RUN echo foo > /foo
 ---> Using cache
 ---> 9c15ea17132c
Step 3/5 : FROM stage1 AS stage2
 ---> 9c15ea17132c
Step 4/5 : RUN echo bar > /bar
 ---> Using cache
 ---> e8adacd0486f
Step 5/5 : COPY baz /
COPY failed: file not found in build context or excluded by .dockerignore: stat baz: file does not exist
```
Podman output:
```
root@ubuntu:~/rebuild# podman build . -q
Error: error building at STEP "COPY baz /": checking on sources under "/root/rebuild": copier: stat: "/baz": no such file or directory
```

So I think still this change is 'correct', i.e. more robust without sacrificing debuggability.

gabrieldemarmiesse:
It's all good for docker then if they still print everything in case of errors. For podman, won't that be annoying to not have the full output?

On a completely independent note I also wonder if podman will one day use buildkit, which would simplify the interfacing we're doing now.

LewisGaul:
> It's all good for docker then if they still print everything in case of errors. For podman, won't that be annoying to not have the full output?

That podman error seems ok to me.

> On a completely independent note I also wonder if podman will one day use buildkit, which would simplify the interfacing we're doing now.

I'm not sure, podman uses `buildah`, which I guess is completely separate?

gabrieldemarmiesse:
Seems ok to me then. We can always talk about it again if someone is complaining about the podman output. Changing this does not necessarily means breaking the api so we'll have flexibility.

I won't make a new release right away, you can always use the fix by pip installing this package from github :)

gabrieldemarmiesse (approved):

LewisGaul:
> I won't make a new release right away, you can always use the fix by pip installing this package from github :) 

Yeah, I've been doing this for testing. No immediate release is not a problem since I was just experimenting with Podman at this point, but I'm reluctant to pin to github in my `requirements.txt`, so in some cases a release is very much appreciated (hence my asking before).

PR #366, opened 2022-09-06
implement docker compose port command

vrazdalovschi:
Add support for `docker compose port` command.

### DIFF
--- a/python_on_whales/components/compose/cli_wrapper.py
+++ b/python_on_whales/components/compose/cli_wrapper.py
@@ -291,14 +291,16 @@ class ComposeCLI(DockerCLICaller):
     def port(
         self, service: str, private_port: str, index: int = 1, protocol: str = "tcp"
     ) -> Tuple[Optional[str], Optional[int]]:
-        """Print the public port for a port binding.
+        """Returns the public port for a port binding.
+
         # Arguments
             service: The name of the service.
             private_port: The private port.
             index: Index of the container if service has multiple replicas (default 1)
             protocol: tcp or udp (default "tcp").
+
         # Returns
-            tuple with [host, port]. If port is unknown, then host and port are None.
+            tuple with (host, port). If port is unknown, then host and port are None.
         """
         full_cmd = self.docker_compose_cmd + ["port"]
         if service == "":
--- a/python_on_whales/components/compose/cli_wrapper.py
+++ b/python_on_whales/components/compose/cli_wrapper.py
@@ -288,9 +288,36 @@ class ComposeCLI(DockerCLICaller):
             full_cmd += to_list(services)
         run(full_cmd)
 
-    def port(self):
-        """Not yet implemented"""
-        raise NotImplementedError
+    def port(
+        self, service: str, private_port: str, index: int = 1, protocol: str = "tcp"
+    ) -> Tuple[Optional[str], Optional[int]]:
+        """Print the public port for a port binding.
+        # Arguments
+            service: The name of the service.
+            private_port: The private port.
+            index: Index of the container if service has multiple replicas (default 1)
+            protocol: tcp or udp (default "tcp").
+        # Returns
+            tuple with [host, port]. If port is unknown, then host and port are None.
+        """
+        full_cmd = self.docker_compose_cmd + ["port"]
+        if service == "":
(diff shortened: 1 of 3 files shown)


vrazdalovschi:
Ops, formatting issue :) I've fixed it and run action in the forked repository.
https://github.com/vrazdalovschi/python-on-whales/runs/8207752632

python_on_whales/components/compose/cli_wrapper.py
```diff
@@ -288,9 +288,34 @@ def pause(self, services: Union[str, List[str], None] = None):
...
+        if private_port == "":
+            raise ValueError("Private port cannot be empty")
+
+        full_cmd.add_simple_arg("--index", index)
+        full_cmd.add_simple_arg("--protocol", protocol)
+        full_cmd += [service, private_port]
+
+        result = run(full_cmd)
+
+        # docker compose could return host:port, remove host from it
```
gabrieldemarmiesse, line 315: Maybe we should return both to stick with what the CLI does? We could return a tuple for example.

gabrieldemarmiesse (commented):
Thank you very much for the pull request! I just have a simple question

python_on_whales/components/compose/cli_wrapper.py
vrazdalovschi: You are right. It would be better to be consistent in the format with CLI. Otherwise, it would be an unexpected result.

gabrieldemarmiesse: I think we should return a tuple to avoid the parsing step for the user.
We should return `("localhost", 8080)` for exemple, or `(None, 8080)` if the CLI doesn't return the host.

With this setup, it will be easier for users to use the result :)

vrazdalovschi: https://github.com/docker/compose/blob/v2/cmd/compose/port.go#L75
`fmt.Printf("%s:%d\n", ip, port)`

heh, so, they join the result, we split it and return tuple?
The `":0"` result for unknown port is caused by default value of type. String default is empty, int default is zero, so it's why their result is `:0`. Would you expect to have `(None, 0:int)` for unknown port and `("host":str, port: int)` for known port?

gabrieldemarmiesse: Ideally, None would mean unknown value. 
It's easier to use in python and leads to less errors down the road for our users as the type is clear.

It's one of those times where we might have to deviate a bit with what the CLI returns to make it easier for our users to use those results.

Of could we should add some documentation in the docstring explaining what None does

So I believe the return type should be `tuple[Optional[str], Optional[int]]`. I'm not super knowledgable with the port command of compose so if what I'm saying doesn't make sense, feel free to tell me 😅

vrazdalovschi: Agreed to return Tuple. I have pushed changes. Let me know if the comment is clear, maybe I should rephrase it

python_on_whales/components/compose/cli_wrapper.py
```diff
@@ -288,9 +288,36 @@ def pause(self, services: Union[str, List[str], None] = None):
...
+        self, service: str, private_port: str, index: int = 1, protocol: str = "tcp"
+    ) -> Tuple[Optional[str], Optional[int]]:
+        """Print the public port for a port binding.
+        # Arguments
+            service: The name of the service.
+            private_port: The private port.
+            index: Index of the container if service has multiple replicas (default 1)
+            protocol: tcp or udp (default "tcp").
+        # Returns
+            tuple with [host, port]. If port is unknown, then host and port are None.
```
gabrieldemarmiesse, line 301:
```suggestion
        """Returns the public port for a port binding.

        # Arguments
            service: The name of the service.
            private_port: The private port.
            index: Index of the container if service has multiple replicas (default 1)
            protocol: tcp or udp (default "tcp").

        # Returns
            tuple with (host, port). If port is unknown, then host and port are None.
```

gabrieldemarmiesse (changes requested):

gabrieldemarmiesse (approved):

gabrieldemarmiesse:
Thank you for the pull request!!!

Issue #367, opened 2022-09-24
docker compose ls -> docker.compose.ls()?


marshall7m:
AFAIK there isn't a method to handle `docker compose ls` CLI command. Just wondering if there's anything in the works to add support for this.

gabrieldemarmiesse:
At first I thought that it was equivalent to `docker.ps()` but it seems that it filters per compose project. Thus it makes sense to implement it. I'll do it when I have time. If you wish you can open a pull request too :)

marshall7m:
@gabrieldemarmiesse thanks for the response. I can get a PR going sometime this week

gabrieldemarmiesse:
That would be awesome, thanks!

PR #368, opened 2022-09-27
docker compose ls -> docker.compose.ls()

marshall7m:
Fixes issue #367 

Implements docker.compose.ls() method that wraps around `docker compose ls` command

I'd be more than happy to make any styling changes and add any additional tests

### DIFF
--- a/python_on_whales/components/compose/cli_wrapper.py
+++ b/python_on_whales/components/compose/cli_wrapper.py
@@ -338,6 +338,21 @@ class ComposeCLI(DockerCLICaller):
         Container = python_on_whales.components.container.cli_wrapper.Container
         return [Container(self.client_config, x, is_immutable_id=True) for x in ids]
 
+    def ls(
+        self, all_stopped=False, project_filters={}
+    ) -> List[python_on_whales.components.compose.cli_wrapper.Projects]:
+        """Returns a list of docker compose projects
+
+        # Returns
+            A `List[python_on_whales.Projects]`
+        """
+        full_cmd = self.docker_compose_cmd + ["ls", "--format", "json"]
+        full_cmd.add_flag("--all", all_stopped)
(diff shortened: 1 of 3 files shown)


gabrieldemarmiesse:
Sorry for the delay of the review, I'm currenly moving, so I might now have to time this week to look at it :(

marshall7m:
> Sorry for the delay of the review, I'm currenly moving, so I might now have the time this week to look at it :(

No worries, thanks for the update. When you do get a chance if you can help me figure out why the `ConfigFiles` key is not available from the `json.loads(run(full_cmd))` output. Strangely, it's available when I run it locally but not within the GH actions job which causes the test to fail on the `ConfigFiles` assertion. It's definitely not a race condition in terms of the project not being available in time for the `docker compose ls` command given the `json.loads(run(full_cmd))` output contains: `[{'Name': 'test_compose_ls', 'Status': 'running(1)'}]` (see [gh job](https://github.com/gabrieldemarmiesse/python-on-whales/actions/runs/3147294506/jobs/5116661874))

gabrieldemarmiesse:
docker compose is in constant evolution. I wouldn't be surprised if you don't have the same version of compose locally and in the CI (the CI probably has an older version). Thus depending on the version we shouldn't assume that the ConfigFiles key will always be present. 

If you still want to test it, and make sure the test works on any platform, you can try to feed your function a json string that was copy-pasted from the output of `docker compose ls` on your computer

Also, can you try to avoid using regex? It's difficult for new contributors to understand the type of parsing that is being done. Using `.split()` should be enough in this situation :)

gabrieldemarmiesse (changes requested):
Very cool pull request, just some small changes and we can merge :)

gabrieldemarmiesse (approved):

Issue #370, opened 2022-10-05
Request for improved handling of decode errors during builds.


Cptmorgan27:
Hi,

Recently, my team encountered an error during a `docker build` where we hit the docker build log limit, and the output was clipped; this resulted in a random non-utf-8 character being produced. The log limit non-deterministically clipped off content in the middle of a UTF-8 Character. In our logs, there was a UTF-8 Character ➤ present; this is made up of 3 bytes (see [Find all Unicode Characters from Hieroglyphs to Dingbats – Unicode Compart](https://www.compart.com/en/unicode/U+27A4) ) 0xE2 0x9E 0xA4. the 1MiB log clipping terminates the output at 0xE2 0x9E. It took us some time and troubleshooting to come to this finding. 

We are asking if there could be more graceful handling of these decoding errors implemented as we believe that the [current handling](https://github.com/gabrieldemarmiesse/python-on-whales/blob/58b52b62949bdbb3a583dbeeac045b34ca0234b5/python_on_whales/components/buildx/cli_wrapper.py#L561-L563) covered up the log limit error and produced a stack trace with the not-so-helpful `UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 10-11: invalid continuation byte`

Thank you!

gabrieldemarmiesse:
Thank you for the issue! Can you tell me how you called buildx exactly? which arguments in the function were used I mean. I'm trying to understand better the situation. Also since it happened to you, what do you think would be a good error message to display to the user in this situation? Thanks!

Cptmorgan27:
We were calling `python_on_whales.docker.build` as such
```python
python_on_whales.docker.build(
                context_path=Path.cwd(),
                secrets=secrets,
                stream_logs=True,
                **docker_build_params.get_denormalized_params(...)
```
We are currently working on adding `builder="larger_log"` to the function call to help with the build log limit.

But yes, something along @LewisGaul comment would do. Mainly a way that allows for the docker build error `[output clipped, log limit 1MiB reached] to be raised to the user.

gabrieldemarmiesse:
Thank you for the feedback! It's much clearer. A pull request is welcomed if someone is motivated, otherwise I'll implement it when I have the time

magichair:
@gabrieldemarmiesse Happy to get a new release cut whenever you're ready and we can test it out over here. 🙏 Thank you.

gabrieldemarmiesse:
[...]

PR #372, opened 2022-10-10
Fix for Issue #370 - Replace on decode errors rather than surface UnicodeDecodeError

magichair:
Fix for https://github.com/gabrieldemarmiesse/python-on-whales/issues/370

Confirmed that this test, without the `errors='replace'` flag, does throw the UnicodeDecodeError. And now, with the change, will do its best effort and replace with the bad output with `�` as most terminals would do.


gabrieldemarmiesse (approved):
Thank you!

Issue #373, opened 2022-10-12
Using hidden 'docker network list' command, which fails on podman


LewisGaul:
https://github.com/gabrieldemarmiesse/python-on-whales/blob/58310714bf768bbb36791b9f8ed8fba7f693fa12/python_on_whales/components/network/cli_wrapper.py#L225

The `ctr_client.network.list()` function is implemented using `<ctr_client> network list`, but this is a hidden (deprecated?) alias of `<ctr_client> network ls` for docker (and podman).
```
root@ubuntu:~# docker network --help

Usage:  docker network COMMAND

Manage networks

Commands:
  connect     Connect a container to a network
  create      Create a network
  disconnect  Disconnect a container from a network
  inspect     Display detailed information on one or more networks
  ls          List networks
  prune       Remove all unused networks
  rm          Remove one or more networks

Run 'docker network COMMAND --help' for more information on a command.
```

This would be *fine*, except podman doesn't seem to properly implement `podman network list`, where the use of `--no-trunc` causes it to fail:
```
root@ubuntu:~# podman network list --no-trunc
Error: unknown flag: --no-trunc
root@ubuntu:~# podman network ls --no-trunc
NETWORK ID                                                        NAME        VERSION     PLUGINS
2f259bab93aaaaa2542ba43ef33eb990d0999ee1b9924b557b7be53c0b7a1bb9  podman      0.4.0       bridge,portmap,firewall,tuning
```

Of course, it could be argued this is a podman bug, but given this is a hidden command it seems unlikely they'll care to fix it (plus I want to be able to use older versions of podman!).

I've confirmed that switching to using `network ls` in this library fixes the issue I'm seeing.

I can open a PR, hopefully you're open to this minor fix (unless I'm missing some context around the use of 'network list'!).

gabrieldemarmiesse:
Thanks you very much for the report and the PR, it shows that we could use some podman tests in the CI as per https://github.com/gabrieldemarmiesse/python-on-whales/issues/353

PR #374, opened 2022-10-12
Switch from 'network list' to 'network ls'

LewisGaul:
Fixes https://github.com/gabrieldemarmiesse/python-on-whales/issues/373


gabrieldemarmiesse (approved):
I don't really mind, those should be equivalent in Docker, if that can increase compatibility let's go for it

PR #377, opened 2022-10-27
Use separate status attributes for ComposeProject model

marshall7m:
The `docker compose ls` status attribute may contain multiple container statuses which can break the current status parsing implementation within the `docker.compose.ls()` method. For example, if the docker project that is associated with the docker client has one container running and another exited, the status will be: `running(1), exited(1)`. 

To handle this case, this PR adds each of the possible statuses as an attribute to the `ComposeProject` model. If the status is not found within the `docker compose ls` output, then the status attribute defaults to `0`.


gabrieldemarmiesse (approved):
Looks good!

PR #378, opened 2022-10-29
Implement `manifest` command

raylas:
Implements the `docker manifest` command as there was only a placeholder in the code.

Official documentation:
https://docs.docker.com/engine/reference/commandline/manifest/

Thanks!


gabrieldemarmiesse (approved):
Super cool thanks!!!

Issue #369, opened 2022-09-27
Feature Request - option to run container rather than pull and run


bthoughton:
Hello,

We are trying to utilize `python_on_whales` in a slightly unusual situation. We created a CLI which implements a subset of Docker commands to execute containers in the Kubernetes, instead of locally. A certain number of docker commands from the CLI of the cluster will (from the perspective of the user) appear to act just like they normally do. Since `python_on_whales` seems to primarily interface with docker through the CLI it is a great fit for us. 

However, we are running into issues when calling `python_on_whales.docker.run()` method, as it calls `docker image inspect` and `docker image pull`. Kubernetes API does not have image management, as a result we cannot implement any of the `docker image ...` commands. 

Ideally, `python_on_whales` would have an optional argument which rather than issuing `docker image` commands would simply perform a `docker run` command, which should work for us and also work in a local Docker environment as well.

Thank you for your time and consideration!

Cheers,

Ben

gabrieldemarmiesse:
Hello, I think in recent version of Docker, there is an argument to docker run that would allow us to do that.

```
    --pull string                    Pull image before running ("always"|"missing"|"never") (default "missing")
```

In your case, if we use "never" then python-on-whales should not try to inspect and pull the docker image. We should implement this feature :)

bthoughton:
That would be great if this was implemented!

gabrieldemarmiesse:
No worries, with the implementation of @MisterOwlPT it should work out and never call `docker image ...` if you specify `pull="never"`.

gabrieldemarmiesse:
Fixed with the new flag `pull="never"` :)

PR #379, opened 2022-10-30
Implementing --pull flag for commands 'run' and 'create' and functions 'copy_from' and 'copy_to' (#369)

MisterOwlPT:
Hi everyone :) 

I found the open issue [#369](https://github.com/gabrieldemarmiesse/python-on-whales/issues/369) and decided to implement the flag _--pull_.

With the proposed changes the following now happens (assume that I don't yet have the desired image stored locally:

```python
>>> from python_on_whales import docker
>>> docker.container.run('alpine:latest', pull='never')  # assumes that image exists locally
docker: Error response from daemon: No such image: alpine:latest.
```

Let's now pull the desired image:

```python

>>> docker.container.run('alpine:latest', pull='missing')  # pulls the image only if required
Unable to find image 'alpine:latest' locally
latest: Pulling from library/alpine
213ec9aee27d: Already exists
Digest: sha256:bc41182d7ef5ffc53a40b044e725193bc10142a1243f395ee852a8d9730fc2ad
Status: Downloaded newer image for alpine:latest
''
```

If now I rerun both the previous commands the containers will start automatically as I already have the desired image stored locally: 

```python

>>> docker.container.run('alpine:latest', pull='never')
''
>>> docker.container.run('alpine:latest', pull='missing')
''
```
If required I can force the image to be pulled every time:

```python

>>> docker.container.run('alpine:latest', pull='always')  # always pulls the image
latest: Pulling from library/alpine
Digest: sha256:bc41182d7ef5ffc53a40b044e725193bc10142a1243f395ee852a8d9730fc2ad
Status: Image is up to date for alpine:latest
''
>>> docker.container.run('alpine:latest', pull='always')
latest: Pulling from library/alpine
Digest: sha256:bc41182d7ef5ffc53a40b044e725193bc10142a1243f395ee852a8d9730fc2ad
Status: Image is up to date for alpine:latest
''
```

By default, the value _missing_ is assumed, identically to the Docker CLI.
In this case, everything works exactly the same as before without any required changes.

```python
>>> docker.container.run('alpine:latest')  # same as docker.container.run('alpine:latest', pull='missing')
```

The same exact mechanism applies to the _create_ command.

```python
>>> docker.container.create('alpine:latest')  # same as  docker.container.create('alpine:latest', pull='missing')
python_on_whales.Container(id='ce4add4deb27', name='confident_davinci')

# The following command takes a bit more time due to pulling the image
>>> docker.container.create('alpine:latest', pull='always')
python_on_whales.Container(id='12584d2d79bb', name='beautiful_bassi')
```

Finally, I think it could be helpful also to control when to pull an image whilst using the functions ```docker.image.copy_from``` and ```docker.image.copy_to```.  Therefore the exact same mechanism was then implemented for these functions.

Suggested changes include unit tests.


python_on_whales/components/container/cli_wrapper.py
```diff
@@ -704,6 +702,8 @@ def create(
         self._add_publish_to_command(full_cmd, publish)
         full_cmd.add_flag("--publish-all", publish_all)
 
+        full_cmd.add_simple_arg("--pull", pull)
```
gabrieldemarmiesse, line 705: the pull was done separately before so that it doesn't appear in stderr and stdout when getting the output of the container. I think we should keep it that way :)

gabrieldemarmiesse (changes requested):

gabrieldemarmiesse:
Looks good otherwise :)

python_on_whales/components/container/cli_wrapper.py
MisterOwlPT: Hi :)

If I understood correctly, the "pull" you are referring to is done by the function `_pull_if_necessary` called inside the bodies of `run` and `create` right? If so, this function _forces_ the behavior of `"missing"` (an image is always pulled if it does not exist locally). It does not support the `"always"` and `"never"` behaviors in any way (hence the open issue).

I had not considered the _stderr_ and _stdout_ related issue (whoops) and I totally agree with you.

Having all this into consideration I've made new changes to the code.
Now, the same new features I described earlier are kept and no image pull message appears in the container  _stream_ (check the following example using `"always"`).

```python
>>> from python_on_whales import docker
>>> stream = docker.run('alpine:latest', command=['ls'], stream=True, pull='always')
latest: Pulling from library/alpine
Digest: sha256:bc41182d7ef5ffc53a40b044e725193bc10142a1243f395ee852a8d9730fc2ad
Status: Image is up to date for alpine:latest
>>> for msg in stream:
...     print(msg)
... 
('stdout', b'bin\n')
('stdout', b'dev\n')
('stdout', b'etc\n')
('stdout', b'home\n')
('stdout', b'lib\n')
('stdout', b'media\n')
('stdout', b'mnt\n')
('stdout', b'opt\n')
('stdout', b'proc\n')
('stdout', b'root\n')
('stdout', b'run\n')
('stdout', b'sbin\n')
('stdout', b'srv\n')
('stdout', b'sys\n')
('stdout', b'tmp\n')
('stdout', b'usr\n')
('stdout', b'var\n')
```

New changes also include updated unit tests for the `run` and `create` functions.

python_on_whales/components/container/cli_wrapper.py
```diff
@@ -737,6 +737,7 @@ def create(
 
         full_cmd.append(image)
         full_cmd += command
+
```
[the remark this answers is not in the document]
MisterOwlPT: Whoops... I am sorry about this. I'll be more careful in future contributions.
Thank you for the advice :)

python_on_whales/components/container/cli_wrapper.py
```diff
@@ -1530,6 +1546,8 @@ def run(
         self._add_publish_to_command(full_cmd, publish)
         full_cmd.add_flag("--publish-all", publish_all)
 
+        full_cmd.add_simple_arg("--pull", "never")
```
gabrieldemarmiesse, line 1549: I think we should only add `--pull never` when the user explicitly ask for it, notably because older docker clients don't have this option and it might render python-on-whales incompatible with those. Other than that, all good :)

python_on_whales/components/container/cli_wrapper.py
```diff
@@ -704,6 +711,8 @@ def create(
...
```
gabrieldemarmiesse, line 714: same here

gabrieldemarmiesse (changes requested):

python_on_whales/components/container/cli_wrapper.py
gabrieldemarmiesse: Very nice analysis thanks :)

MisterOwlPT:
**Problem**:

As you said, if `--pull` _is_ forced then compatibility with older client versions is compromised.

If `--pull "never"` _is not_ forced then the Docker image will always be pulled at least once, as the default value is `--pull "missing"`. This:
- Compromises support of `--pull` for newer client versions and may even lead to undesired behaviour;
- Possibility of polluting _stdout_ whenever `docker.run`/`docker.create` is called with argument `pull="never"`.

**Proposal:**

I may be overthinking but I think that in this case some logic _based on the client version_ is required.
Checking the [Docker CLI release notes](https://docs.docker.com/engine/release-notes/) I found that support for `--pull` was added on release `20.10.0`.

Then I implemented a mechanism that checks if the Docker client in use supports `--pull` based on its version.

If the client _supports_ `--pull` (version >= `20.10.0`) the logic within `run`/`create` is kept the same as previously explained:
- Images are pulled according to the value of argument `pull`.
- Images are always pulled before the call to `run` as to not pollute _stdout_;
- `--pull "never"` is _always_ added to `full_cmd` for the previously explained reason.

If the client _does not support_ `--pull` (version < `20.10.0`) the logic within `run`/`create` is the following:
- Argument `pull` is completely ignored (only default value `"missing"` is supported);
- Docker images are always pulled using existing function `_pull_if_required` before the call to `run` as to not pollute _stdout_;
- `--pull` is _never_ added to `full_cmd` for compability reasons.

I'm currently retrieving the Docker client version by running the following command `docker version --format "{{ json .Client.Version}}"`.
A whole ton of other information can be obtained with this command but I'm currently only extracting what I need (I left a _TODO_ for maybe a future contribution).
I placed this logic inside the _DockerCLICaller_ class as to make the extracted information accessible in many places where version-based logic may prove useful.

What do you think of this? :) 

Unit tests were updated to reflect all these changes.

LewisGaul:
I'm not convinced by the idea of checking the client version (is the daemon version relevant too? what about podman or other drop-in runtimes?). My understanding is that this package is meant to be a thin wrapper around whatever client you have, meaning the defaults should be the same as with the CLI, so I agree with
> I think we should only add `--pull never` when the user explicitly ask for it

MisterOwlPT:
Hi :)

After rereading your comments.... oh boy was I overthinking and complicating things!

I reverted the previous changes and I'm now ensuring that `--pull "never"` is only added to `full_cmd`
if the user specifies it, as you both have suggested, without any complicated logic.
The CLI defaults are preserved and the _stdout_ logs are kept clean.

It is up to users to properly use `python_on_whales` according to the version installed on their machines.

Unit tests also reverted and updated for the proposed changes.

python_on_whales/components/compose/cli_wrapper.py
```diff
@@ -354,16 +353,11 @@ def ls(
...
         return [
             ComposeProject(
                 name=proj["Name"],
-                created=parse_ls_status_count(proj["Status"], "created"),
-                running=parse_ls_status_count(proj["Status"], "running"),
-                restarting=parse_ls_status_count(proj["Status"], "restarting"),
-                exited=parse_ls_status_count(proj["Status"], "exited"),
-                paused=parse_ls_status_count(proj["Status"], "paused"),
-                dead=parse_ls_status_count(proj["Status"], "dead"),
+                status=proj["Status"].split("(")[0],
```
gabrieldemarmiesse, line 359: I believe there was an issue with the rebase. I advise using merge when in doubt :)

gabrieldemarmiesse (changes requested):

PR #380, opened 2022-11-06
Add envs parameter to compose up/config/run

lugoues:
Docker compose allows variable injection from the process environment in addition to a `.env` file. This change will allow user the ability to pass a dictionary of variables to be injected into the compose subprocess.

### DIFF
--- a/python_on_whales/client_config.py
+++ b/python_on_whales/client_config.py
@@ -64,6 +64,7 @@ class ClientConfig:
     compose_project_name: Optional[str] = None
     compose_project_directory: Optional[ValidPath] = None
     compose_compatibility: Optional[bool] = None
+    compose_envs: Optional[Dict[str, str]] = None
     client_call: List[str] = field(default_factory=lambda: ["docker"])
     _client_call_with_path: Optional[List[Union[Path, str]]] = None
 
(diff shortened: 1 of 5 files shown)


gabrieldemarmiesse:
Thanks for the pull request, that parameter should be added to the client constructor, similarely to other compose parameters like the file path or project name. Thanks!

Lugoues:
> Thanks for the pull request, that parameter should be added to the client constructor, similarely to other compose parameters like the file path or project name. Thanks!

I had avoided adding it to the client as it is not a flag and all the supporting code only deals with flags.  The method I would take to implement this would be:
1. add `compose_envs` to `DockerClient` constructor
2. add `compose_envs` to `ClientConfig`
3. add `envs` to all `run` statements, pulling from `self.client_config.compose_envs` within the `ComposeCLI` class.

This begs the question, would we want to support the `envs` for every subprocess call in the client? The container class has several calls where it accepts `envs` that we could set at the client level.

gabrieldemarmiesse:
Similar to how we use "self.docker_compose_cmd" in every call to compose, we should use "self.docker_compose_envs". Because compose commands parse the compose file to understand what is available and how to behave with services, we should passe it to all commands. For a better user experience, we wouldn't want the user to specify those environement variables every time. Just once in the DockerClient constructor seems fine :) Similar to all the other compose flags

timini:
any news on this?

Lugoues:
Hey, yeah. I have most of it done but life/work he's eating my time. I hope to
finish it soon. Sorry for the delay!

Tardo:
Thanks for the work... any news?

gabrieldemarmiesse:
Closing as stale. Feel free to open another PR if you want!

PR #381, opened 2022-11-07
Implementing --pull flag for commands 'run' and 'create' and functions 'copy_from' and 'copy_to' (#369)

MisterOwlPT:
Hi :) 

I am opening this pull request as a continuation of "old" pull request #379.

As pointed out,  I created many issues while during rebase (whoops...). 
I found it easier to do some cherry-picking and open this new pull request.
I hope this solution is okay for you :)

All issues are now resolved and no change was made to the previous changes.

### DIFF
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -747,7 +747,6 @@ class ContainerCLI(DockerCLICaller):
 
         full_cmd.append(image)
         full_cmd += command
-
         return Container(self.client_config, run(full_cmd), is_immutable_id=True)
 
     def _add_publish_to_command(self, full_cmd, publish):
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -175,6 +175,23 @@ class Container(ReloadableObjectFromJson):
     # --------------------------------------------------------------------
     # public methods
 
+    def attach(
+        self,
+        detach_keys: Optional[str] = None,
+        no_stdin: bool = False,
+        sig_proxy: bool = True
+    ) -> Union[None, str, Iterable[Tuple[str, bytes]]]:
+        """Attach local standard input, output, and error streams to a running container.
+
+        Alias: `docker.attach(...)`
+
+        See the [`docker.container.attach`](../sub-commands/container.md#attach) command for
+        information about the arguments.
+        """
+        return ContainerCLI(self.client_config).attach(
+            self, detach_keys, no_stdin, sig_proxy
+        )
+
     def commit(
         self,
         tag: Optional[str] = None,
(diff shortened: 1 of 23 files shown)


python_on_whales/components/container/cli_wrapper.py
```diff
@@ -737,6 +747,7 @@ def create(
 
         full_cmd.append(image)
         full_cmd += command
+
         return Container(self.client_config, run(full_cmd), is_immutable_id=True)
```
gabrieldemarmiesse, line 751:
```suggestion
        full_cmd += command
        return Container(self.client_config, run(full_cmd), is_immutable_id=True)
```

tests/python_on_whales/components/test_container.py
```diff
@@ -702,3 +702,170 @@ def test_attach_sig_proxy_argument(run_mock: Mock, inspect_mock: Mock) -> None:
     run_mock.assert_called_once_with(
         docker.client_config.docker_cmd + ["attach", test_container_name], tty=True
     )
+
+
+@patch("python_on_whales.components.container.cli_wrapper.run")
```
gabrieldemarmiesse, line 707: For the unit tests, I would prefer if we don't use `patch` unless it's absolutely necessary.  You can try for exemple to test the "never" option by removing the docker image first and do a docker run after, then check that an exception has been raised.

To check the "missing" you can test by using docker run when the docker image is not present locally, it should not crash. When the docker image is present locally, it should not try to get an updated version.

You can update a docker image in a registry by using the fixture that launch a registry locally. Then you'll have rights to pull and push to it. Take a look at the conftest.py

for the "always", we can check that it works by pulling the image first, pushing an updated version to the registry and then trying to do docker run again. It should have updated the docker image locally.

It's always better to avoid `patch` if we can, because then not all the code is executed in the unit test

gabrieldemarmiesse (changes requested):

tests/python_on_whales/components/test_container.py
MisterOwlPT: Actually, I was actually using `patch` on purpose so as to not execute all the code.

Since `python_on_whales` is a wrapper for the Docker CLI, why isn't it enough to just ensure that the CLI is called with the proper arguments for each case? Testing `python_on_whales` by actually calling Docker and analyzing its outcomes sounds to me like not only the wrapper is being tested but also Docker itself (up to some extent).

I think that makes sense for functions such as `copy_to` and `copy_from` that are not part of the CLI itself.
It may also make sense when testing the function `utils.run` since it calls the actual CLI.

However, in this case, I think it suffices to ensure that the expected arguments are passed to the CLI and that the images are pulled as expected (without having to actually pull them).

Let me know your thoughts on this :)

gabrieldemarmiesse: Well in theory, yes, in practice, we might make an error in the expected command in the unit tests. Let's say that you make a unit test and you say with patch that you expect the command called to be 
`docker run -v /tmp:/tmp,/somedir:/some_other_dir ubuntu`. Then the test will pass because you thought this was the right syntax, but by running the real code, you would get an error and thus correct the wrong code. I had to do some trial and error when making this package because the docker CLI documentation is not always clear, so I would prefer that the real code is running to avoid this kind of issues

MisterOwlPT:
Hey hi :) 

> ... test will pass because you thought this was the right syntax, but by running the real code, you would get an error and thus correct the wrong code.

I see what you mean and agree with it to some extent.
I think, however, that developers should not blindly implement tests for bits of code that never worked in the first place. 

Still, I too think it's better to be safe than sorry.
I followed your previous suggestions (thank you!) and implemented additional unit tests that actually call the CLI and inspect the results, for all possible values of `pull` and for both `create`/`run`.

If you agree, I would suggest keeping the tests using `patch` as I think they can still provide some value.

Let me know what you think :)

gabrieldemarmiesse (approved):
Really awesome tests :)  Indeed I agree with everything you say. I just can't expect everyone who contributes to be as diligent as you concerning verifications in unit tests mocks

Mocks are certainly useful here as we can run tests much faster than in tests that use the real docker engine :) I like to have both if possible. If not, I'm satisfied with the tests that call the docker engine too.

PR #382, opened 2022-11-17
Add a new dummy file (test PR)

### LINKED ISSUE #353: Run some unit tests with podman in the CI
In our CI we only test python-on-whales with the Docker client. While we won't bend our code to adapt to podman, we should avoid any breakage if it's low effort. To ensure this, we should have some of our tests be executed with both podman and docker.

Ideally we should have a decorator/pytest fixture that we could declare for some specific unit tests and it would run with podman and docker  as two different tests. A bit like `@pytest.mark.parametrize()`.

Note that for this to work, podman should be installed and working in the CI.

### LINKED ISSUE #369: Feature Request - option to run container rather than pull and run

gabrieldemarmiesse:
Fix #353 

Fix #369


gabrieldemarmiesse:
Fix #369

PR #383, opened 2022-11-17
[IMP] "docker compose ps" argument services

fuentes73:
Hi everyone!

This is useful if you have many services and you want to choose from which one to list its containers.

From the [docs](https://docs.docker.com/engine/reference/commandline/compose_ps):  ` $ docker compose ps [OPTIONS] [SERVICE...]`


gabrieldemarmiesse (approved):
Many thanks!

PR #385, opened 2022-11-18
Fixed command for removing node labels

### LINKED ISSUE #384: Wrong option used for removing node labels
Hi, I think the wrong option is currently used to remove labels from nodes, causing an exception:

Pasted Graphic

Looks like the correct naming is `--label-rm`:

Update a node

I've checked the option's name in the docker cli repo and doesn't look like it has changed recently - https://github.com/docker/cli/blame/master/cli/command/node/update.go

I will raise a PR to correct it, if that's ok?

Thanks

NickG158:
Fixes #384

### DIFF
--- a/python_on_whales/components/node/cli_wrapper.py
+++ b/python_on_whales/components/node/cli_wrapper.py
@@ -238,7 +238,7 @@ class NodeCLI(DockerCLICaller):
             full_cmd += ["--label-add", f"{label_name}={label_value}"]
 
         for label in rm_labels:
-            full_cmd += ["--rm-label", label]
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse:
Thank you very much for the fix. I think there was a bug because I forgot to add a unit test. Can you add one? It should be easy to add and remove a label on a node in the unit tests I think, if you look around we already test some other swarm function.

NickG158:
I've added tests for adding & removing a label, let me know if they're not up to scratch :)

gabrieldemarmiesse (approved):
Very cool thanks!

PR #386, opened 2022-11-18
[IMP] docker compose run & up

fuentes73:
Hi!

### 1st commit

Add or override default labels.
` docker compose run --label="traefik.enable=false" MY_SERVICE ARG` 
[Docs](https://docs.docker.com/engine/reference/commandline/compose_run/)

###  2nd commit

If containers already exist, don't recreate them. 
`docker compose up --no-recreate --scale MY_SERVICE=2`
[Docs](https://docs.docker.com/engine/reference/commandline/compose_up/)

### DIFF
--- a/python_on_whales/components/compose/cli_wrapper.py
+++ b/python_on_whales/components/compose/cli_wrapper.py
@@ -653,6 +653,7 @@ class ComposeCLI(DockerCLICaller):
         scales: Dict[str, int] = {},
         attach_dependencies: bool = False,
         force_recreate: bool = False,
+        recreate: bool = True,
         no_build: bool = False,
         color: bool = True,
         log_prefix: bool = True,
@@ -682,6 +683,8 @@ class ComposeCLI(DockerCLICaller):
             attach_dependencies: Attach to dependent containers.
             force_recreate: Recreate containers even if their configuration and image
                 haven't changed.
+            recreate: Recreate the containers if already exist.
+                `recreate=False` and `force_recreate=True` are incompatible.
             no_build: Don't build an image, even if it's missing.
             color: If `False`, it will produce monochrome output.
             log_prefix: If `False`, will not display the prefix in the logs.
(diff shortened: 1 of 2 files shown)


tests/python_on_whales/components/test_compose.py
```diff
@@ -469,6 +479,20 @@ def test_compose_run_detach():
     assert container.logs() == "dodo\n"
 
 
+def test_docker_compose_run_labels():
+    container = docker.compose.run(
+        "alpine",
+        ["echo", "dodo"],
+        labels={"traefik.enable": "false"},
```
gabrieldemarmiesse, line 486: Can you also add a test with more than one label? To make sure the syntax is ok.

gabrieldemarmiesse (changes requested):
Thanks for the pull request! Just a small change if you have the time :)

tests/python_on_whales/components/test_compose.py
fuentes73: Done!

gabrieldemarmiesse (approved):
Many thanks for the new flags, a lot of users will appreciate :)

python_on_whales/components/compose/cli_wrapper.py
```diff
@@ -651,6 +653,7 @@ def up(
         scales: Dict[str, int] = {},
         attach_dependencies: bool = False,
         force_recreate: bool = False,
+        no_recreate: bool = False,
```
gabrieldemarmiesse, line 656: Thinking about it a bit longer, the fact that the parameter name is negative (with "no" inside) is a constraint on the command line. In python-on-whales, in other parts of the codebase, we try to make the parameters name positive to be more user friendly.
I know this seems weird because there is "no_build" just below, but i'll deprecate it in the future in favor of `build: bool =True`. 
```suggestion
        recreate: bool = True,
```

gabrieldemarmiesse (changes requested):

python_on_whales/components/compose/cli_wrapper.py
fuentes73: Sure no problem. I force push the changes, let me know what you think. Many thanks

gabrieldemarmiesse: Feel free to organise the commits however you want, I squash them anyway when I merge in master

gabrieldemarmiesse (approved):
Many thanks!

PR #388, opened 2022-11-24
Added support for --remove-orphas docker compose up flag

bodo-ata:
Hi,

Would like to add support for this flag also for docker compose up (was implemented already for down).

When you re-run docker compose up on a file from which services were removed you can see messages like this one:
```
time="2022-11-24T14:56:17+02:00" level=warning msg="Found orphan containers ([tmp_busybox2_1]) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up."
```

The containers for the services removed from the compose file are still running

There are cases when one would want them shutdown/removed instead, hence adding the support for  `--remove-orphans`

### DIFF
--- a/tests/python_on_whales/components/test_compose.py
+++ b/tests/python_on_whales/components/test_compose.py
@@ -807,30 +807,28 @@ services:
         file.write(base_cfg + service_to_remove)
 
     docker.compose.up(detach=True)
-    containers = docker.compose.ps()
-    assert len(containers) == 2
-    container_ids = [container.id for container in containers]
+    compose_containers = docker.compose.ps()
+    assert len(compose_containers) == 2
+    compose_container_ids = {container.id for container in compose_containers}
 
     # updating the docker compose file to have only 1 service configured
     with open(compose_file, "w") as file:
         file.write(base_cfg)
 
-    docker.compose.up(detach=True)
-    containers = docker.compose.ps()
-    assert len(containers) == 1
-    container_ids.remove(containers[0].id)
-    # remaining id belongs to the container of the removed service
-    removed_service_container_id = container_ids[0]
+    def check_running_containers(expected):
+        containers = docker.ps()
+        container_ids = {container.id for container in containers}
+        assert len(compose_container_ids.intersection(container_ids)) == expected
 
-    # container of the removed service is still running
-    assert len(docker.ps(filters={"id": removed_service_container_id})) == 1
+    docker.compose.up(detach=True)
+    # both containers running
+    check_running_containers(2)
 
     # calling with remove_orphans flag
     docker.compose.up(detach=True, remove_orphans=True)
-    assert len(docker.compose.ps()) == 1
-
-    # container of the removed service was stopped
-    assert len(docker.ps(filters={"id": removed_service_container_id})) == 0
+    # orphan container (of the removed service) was stopped
+    check_running_containers(1)
 
     docker.compose.down(timeout=1)
+    check_running_containers(0)
     remove(compose_file)
--- a/tests/python_on_whales/components/test_compose.py
+++ b/tests/python_on_whales/components/test_compose.py
@@ -834,4 +834,3 @@ services:
 
     docker.compose.down(timeout=1)
     remove(compose_file)
-
--- a/tests/python_on_whales/components/test_compose.py
+++ b/tests/python_on_whales/components/test_compose.py
@@ -784,11 +784,11 @@ def test_compose_ls_project_multiple_statuses():
 
 
 def test_docker_compose_up_remove_orphans():
-    compose_file = tempfile.mktemp(prefix='test_docker_compose_up_remove_orphans_', suffix='.yml')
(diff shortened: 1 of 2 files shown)


tests/python_on_whales/components/test_compose.py
```diff
@@ -780,3 +781,56 @@ def test_compose_ls_project_multiple_statuses():
         assert sorted(project.config_files) == sorted(d.client_config.compose_files)
 
     d.compose.down(timeout=1)
+
+
+def test_docker_compose_up_remove_orphans():
```
gabrieldemarmiesse, line 786: Thanks for the PR, you can only have orphans if you change the docker-compose.yml file and remove services in this file as far as I know. Maybe that's why you have issues here in the unit tests?

gabrieldemarmiesse:
Very cool feature flag to have :)

tests/python_on_whales/components/test_compose.py
bodo-ata: Hi @gabrieldemarmiesse, that is a fast response, thank you!

This is exactly what I'm doing in that test, running docker compose up on a file with 2 services,
rewrite the file to include only one, re-run docker compose up, and the re-run docker compose up again using the new flag.

Currently I'm a bit confused why there are differences between my end and git env.
It looks like on my environment once I modify the file and run up again, docker.compose.ps() returns the container only for the remaining service, while for github actions env it returns the container for both services.

Here is what I'm using:
```
$ docker -v
Docker version 20.10.12, build 20.10.12-0ubuntu2~20.04.1
$ docker compose version
Docker Compose version v2.12.2
```
any idea/hint is welcome :)

I'll investigate more what happens actually, and get back with updates.

gabrieldemarmiesse: It's quite possible that the version of compose you are using is not the same as ghe one in the CI. I think the once in the ci is much older

gabrieldemarmiesse: Feel free to upgrade it y'if you need

bodo-ata:
Hi @gabrieldemarmiesse,

I was able to re-write the test not to depend on that difference
could you check if there is anything else needed ?

Thank you!

tests/python_on_whales/components/test_compose.py
```diff
@@ -780,3 +781,54 @@ def test_compose_ls_project_multiple_statuses():
         assert sorted(project.config_files) == sorted(d.client_config.compose_files)
 
     d.compose.down(timeout=1)
+
+
+def test_docker_compose_up_remove_orphans():
+    compose_file = tempfile.mktemp(
+        prefix="test_docker_compose_up_remove_orphans_", suffix=".yml"
+    )
+    docker = DockerClient(
```
gabrieldemarmiesse, line 790:
```suggestion
    )
    compose_file = Path(compose_file)
    docker = DockerClient(
```

tests/python_on_whales/components/test_compose.py
```diff
@@ -780,3 +781,54 @@ def test_compose_ls_project_multiple_statuses():
...
+    command: sleep infinity
+"""
+    service_to_remove = """  busybox2:
+    image: busybox:latest
+    command: sleep infinity
+"""
+
+    # writing the docker compose file with 2 services configured
+    with open(compose_file, "w") as file:
+        file.write(base_cfg + service_to_remove)
```
gabrieldemarmiesse, line 807:
```suggestion
    compose_file.write_text(base_cfg + service_to_remove)
```

tests/python_on_whales/components/test_compose.py
```diff
@@ -780,3 +781,54 @@ def test_compose_ls_project_multiple_statuses():
...
+
+    docker.compose.up(detach=True)
+    compose_containers = docker.compose.ps()
+    assert len(compose_containers) == 2
+    compose_container_ids = {container.id for container in compose_containers}
+
+    # updating the docker compose file to have only 1 service configured
+    with open(compose_file, "w") as file:
+        file.write(base_cfg)
```
gabrieldemarmiesse, line 816:
```suggestion
    compose_file.write_text(base_cfg)
```

tests/python_on_whales/components/test_compose.py
```diff
@@ -780,3 +781,54 @@ def test_compose_ls_project_multiple_statuses():
...
+
+    def check_running_containers(expected):
```
gabrieldemarmiesse, line 818: Please avoid nested functions, you can put it somewhere else in the file.

gabrieldemarmiesse (changes requested):
Please use Pathlib and avoid inner function to improve readability

bodo-ata:
Good feedback @gabrieldemarmiesse, thank you!
Pls let me know if there is something else to adjust.

gabrieldemarmiesse (approved):
Many thanks! It's a great feature :)

bodo-ata:
Thank you @gabrieldemarmiesse , we love this module!

PR #390, opened 2022-12-01
feat: Support for using docker-compose binary instead of docker compose add-in

tmeckel:
This PR contains changes which adds support to use the `docker-compose` binary instead of the `docker compose` add-in. 
This also allows phython-on-whales to be used with podman, since podman itself does not provide compose support.

### DIFF
--- a/python_on_whales/client_config.py
+++ b/python_on_whales/client_config.py
@@ -107,9 +107,13 @@ class ClientConfig:
         )
 
     def _get_docker_compose_path(self) -> List[str]:
-        if 0 == subprocess.call(["docker", "compose"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL):
+        if 0 == subprocess.call(
+            ["docker", "compose"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL
+        ):
             return ["docker", "compose"]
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
Thank you very much for the pull request. I appreciate that you've taken the time to add this feature but sadly, I do not want to support docker compose v1. While it's possible that there will be no issues with v1 in the future, I we declare that python-on-whales supports compose v1, we might get quite a lot issues related to this. 

What we can do though, is to make a note in the readme saying that if users want compose v1 support, they are free to use your fork, but then it would be your responsability to maintain it.

I hope you're not too disapointed with my answer. I made this package because I knew the maintenance would be extremely low (and I think it worked well when I compare that to the maintenance costs of docker-py). I don't want increase this maintenance cost. This is to ensure I'm always quick to answer issues and pull requests :(

tmeckel:
Are you aware of the fact that Docker has rewritten docker-compose in Go and this version has noting to do with the old v1 version? Even more this implementation can be installed separately aside Docker itself. And as I stated I use this setup with podman to use Docker Compose in rootless container setup. 

https://github.com/docker/compose

gabrieldemarmiesse:
Sure, i'm familiar with the project. I'm aware that it means python-on-whales can't control compose and podman at the same time. I don't want to support docker-compose as it will be deprecated soon too

PR #394, opened 2022-12-21
Add docker context create command

toxicphreAK:
The [docker context create](https://docs.docker.com/engine/reference/commandline/context_create/) command is implemented like stated in the official docker docs.


python_on_whales/components/context/cli_wrapper.py
```diff
@@ -81,9 +81,40 @@ def use(self) -> None:
...
+                full_cmd += ["--from", from_]
+            else:
+                full_cmd += ["--docker"]
+                if from_host:
+                    if from_host in existing_contexts:
+                        full_cmd += [f"from={from_host}"]
+                else:
+                    full_cmd += [f"host={host}"]
+                if description:
+                    full_cmd += [f"--description={description}"]
```
gabrieldemarmiesse, line 114: For this kind of situation, can you use the methods already declared in the class? You shouldn't have to manipulate lists at all here, it should make it easier. You can take a look in `docker.container.run()` https://github.com/gabrieldemarmiesse/python-on-whales/blob/master/python_on_whales/components/container/cli_wrapper.py#L611 to have examples of such methods :) 
In this situation I think you can use `self.add_simple_arg()`.

python_on_whales/components/context/cli_wrapper.py
```diff
@@ -81,9 +81,40 @@ def use(self) -> None:
...
+
+        # Arguments
+            context: name of the context to create
+            host: uri of the host to access
+            description: description of the context
+            from_: create context from a named context
+            from_host: use endpoint configuration from existing named context
+        """
+        existing_contexts = [str(x) for x in self.list()]
+        if context not in existing_contexts:
```
gabrieldemarmiesse, line 102: I don't think we should add more logic than what is done in the command line. If the user input will trigger an error, let the error from the command line pop up and it will be sent back to the user as an exception

gabrieldemarmiesse (changes requested):
Thanks you for the new feature :) could you add unit tests with it? thanks!

python_on_whales/components/context/cli_wrapper.py
toxicphreAK: Will change that ones. Thank you and sorry that I am not really into your code.

gabrieldemarmiesse: No worries!

toxicphreAK: The `self.add_simple_arg()` is joining the arguments via whitespace, but I need it with an equal sign. Let me do an example:
`--description test` should become `--description=test`. Can you help me out here?

gabrieldemarmiesse: I think whitespace is ok no? Docker usually accepts whitespace and equal.

toxicphreAK: No actually it runs into an error:
`python_on_whales.exceptions.DockerException: The docker command executed was `/usr/bin/docker context create --docker host ssh://test@test.domain --description Python on whales testing context testpow``

toxicphreAK: It was the combination of `--docker`, `host` and then `value`... will change it.

gabrieldemarmiesse (approved):

gabrieldemarmiesse:
Sorry for the delay, it was more complicated than I anticipated. It should be all good now. I will make a new release once it's merged in master

PR #397, opened 2023-01-12
Add `__all__` to specify what `*` should import from the main module

rafrafek:
I think it is a common practice, but I'm not an expert. Feel free to accept/deny or suggest changes.

### DIFF
--- a/python_on_whales/__init__.py
+++ b/python_on_whales/__init__.py
@@ -22,3 +22,25 @@ from .exceptions import DockerException
 
 # alias
 docker = DockerClient()
+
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse (approved):
Thanks! Looks good to me!

gabrieldemarmiesse:
I'm not sure I understand why the build is failing. I'll look into it. Thanks for your patience :)

rafrafek:
Thanks! No need to hurry :)

I have no idea what is happening with this build...

PR #399, opened 2023-01-16
`return_stderr` should be True for `run` method to return tuple

rafrafek:
I think it's intended to work that way

### DIFF
--- a/python_on_whales/utils.py
+++ b/python_on_whales/utils.py
@@ -75,12 +75,12 @@ class DockerCamelModel(pydantic.BaseModel):
 @overload
 def run(
     args: List[Any],
-    capture_stdout: bool = True,
-    capture_stderr: bool = True,
-    input: bytes = None,
-    return_stderr: Literal[True] = True,
-    env: Dict[str, str] = {},
-    tty: bool = False,
+    capture_stdout: bool = ...,
+    capture_stderr: bool = ...,
+    input: bytes = ...,
+    return_stderr: Literal[True] = ...,
+    env: Dict[str, str] = ...,
+    tty: bool = ...,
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
Maybe there is a misunderstanding? When using overloads, the things that are used in the signature are the types, not the default values. I think the default value is not used here by type checkers, only the default value of the original signature counts

rafrafek:
For me Pylance complains when `return_stderr: Literal[True] = False`:
```
Expression of type "Literal[False]" cannot be assigned to parameter of type "Literal[True]"
  "Literal[False]" cannot be assigned to type "Literal[True]"
```

It makes sense for me to not assign `Literal[False]` to `Literal[True]` 😅

> When using overloads, the things that are used in the signature are the types, not the default values.

I've pushed commit that replaces default values with `...` to address this.

rafrafek:
There is `input: bytes = None` in `run()`, shouldn't it be `input: Optional[bytes] = None`?

https://stackoverflow.com/a/49724148/9737275

That's what I've found:

> Type checkers should move towards requiring the optional type to be made explicit.

gabrieldemarmiesse:
I agree that if the default value is None, then it should be optional. I'll take a look at the docs for the overloads

gabrieldemarmiesse (approved):
Thank you for the pull request!

gabrieldemarmiesse:
> There is input: bytes = None in run(), shouldn't it be input: Optional[bytes] = None?

That is indeed the case. Can you open a new pull request for this fix?

rafrafek:
> That is indeed the case. Can you open a new pull request for this fix?

Done! :)

PR #400, opened 2023-01-18
Add support for the --cgroupns arg

LewisGaul:
See https://docs.docker.com/engine/reference/commandline/create/#options.

This arg was added in https://github.com/docker/cli/pull/2024, v20.10.0 ([link](https://docs.docker.com/engine/release-notes/#client-7)), and is also [supported by podman](https://docs.podman.io/en/latest/markdown/podman-run.1.html#cgroupns-mode).

I've tested this works manually, unsure if you think automated testing should somehow be added too?

### DIFF
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -492,6 +492,7 @@ class ContainerCLI(DockerCLICaller):
         cap_add: List[str] = [],
         cap_drop: List[str] = [],
         cgroup_parent: Optional[str] = None,
+        cgroupns: Optional[str] = None,
         cidfile: Optional[ValidPath] = None,
         cpu_period: Optional[int] = None,
         cpu_quota: Optional[int] = None,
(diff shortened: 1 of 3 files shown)


gabrieldemarmiesse:
Thanks a lot for the new feature! If it's possible to make a small test that would be the best. If the option can be seen with just a docker inspect on the container, that'd be great :)

LewisGaul:
Good to be back with a contribution to one of my favourite open source projects!

I've added the `cgroupns_mode` field to the `ContainerHostConfig` (thanks for pointing that out). My concern is that if using a version of docker before v20.10.0 (or around then) when support for cgroupns was added, then will this give an error or will the field be set to 'None'? (I made the field `Optional[str]` in hope of the latter). I can test this out manually today.

gabrieldemarmiesse:
You don't need the optional. This is because a decorator already put Optional around all types of the model automatically to ensure backward compatibility :)

LewisGaul:
The linux tests seem to be failing running `wget` on `https://github.com/docker/compose/releases/download/v2.2.3/docker-compose-linux-x86_64` for some reason...

gabrieldemarmiesse:
Yes, I think something was upgraded automatically in the ci. I'll take a look at it this weekend and I'll fix it. Sorry for the delay!

LewisGaul:
The CI passed on my other PR - are you able to retrigger it on this PR?

gabrieldemarmiesse:
This is a very strange error, I'm quite confused

gabrieldemarmiesse (approved):

LewisGaul:
I think the CI issue might've been caused by the test I added (container sleeps forever but is started attached!), hopefully I've fixed it now!

gabrieldemarmiesse:
Nice! Thanks a lot! I think I'll make a new release once #407 has been fixed

Issue #434, opened 2023-04-13
Use another library to generate our documentation


gabrieldemarmiesse:
We currently use keras-autodoc https://github.com/keras-team/keras-autodoc to generate the markdown for our documentation. We then use mkdocs to convert that to html.

Keras-autodoc isn't supported anymore, we need a remplacement before the CI starts breaking (shouldn't be urgent but we need to do it at some point). We should look into a replacement like https://github.com/NiklasRosenstein/pydoc-markdown or similar https://github.com/mkdocstrings/mkdocstrings

It doesn't matter if the documentation is different afterwards

kclamalex:
Hey I could take a look at it! Will prepare the PR for this.

gabrieldemarmiesse:
That would be great, thanks! Don't hesitate to ask if you need help or guidance!

gabrieldemarmiesse:
It looks really good! Thanks @kclamalex !

PR #444, opened 2023-05-24
Use mkdocstrings to generate documentation

kclamalex:
Fixes https://github.com/gabrieldemarmiesse/python-on-whales/issues/434

As I can see [pydoc-markdown](https://github.com/NiklasRosenstein/pydoc-markdown) is not actively maintained anymore and don't want to switch library in the near future, I guess it's safer to use mkdocstrings.

However, there are already some known issues when I use mkdocstrings:
1. Seems it doesn't provide the function to replace signature, at least I couldn't find one. Therefore, the signature is showing the original class name, instead of the desired signature (e.g. instead of "docker.buildx.",  "BuildxCLI" is shown )
2. Attribute is showing in code block, instead of in bullet points.

Other than that, I couldn't find other differences at the moment.


docs/autogen.py
```diff
@@ -1,147 +1,128 @@
...
+    pages: t.Dict[str, list], template_path: str, destination_path: str
+) -> None:
+    for md_file in pages:
+        template_file_path = f"{template_path}/{md_file}"
+        target_file_path = f"{destination_path}/{md_file}"
+        os.makedirs(Path(target_file_path).parent, exist_ok=True)
 
+        if Path(template_file_path).exists():
+            with open(template_file_path, "r") as fin:
+                filedata = fin.read()
```
gabrieldemarmiesse, line 91:
```suggestion
            filedata = template_file_path.read_text()
```

docs/autogen.py
```diff
@@ -1,147 +1,128 @@
+import os
 import shutil
+import typing as t
```
gabrieldemarmiesse, line 3:
```suggestion
import typing
```
Same

gabrieldemarmiesse (changes requested):
Thank you for this awesome work!!! I will generate the doc locally to see what it looks like and the give some feedback :)

docs/autogen.py
```diff
@@ -1,147 +1,126 @@
...
-        "python_on_whales.Stack",
-        "python_on_whales.Task",
-        "python_on_whales.Volume",
-    ],
-    titles_size="##",
-)
+def md_generate(
+    pages: Dict[str, list], template_path: str, destination_path: str
+) -> None:
+    for md_file in pages:
```
gabrieldemarmiesse, line 84:
```suggestion
    for md_file, module_list in pages.items():
```

docs/autogen.py
```diff
@@ -1,147 +1,126 @@
...
+            filedata = AUTOGEN_PATTERN
 
-destination = PROJECT_ROOT / "docs" / "generated_sources"
-doc_generator.generate(destination)
-shutil.copyfile(PROJECT_ROOT / "README.md", destination / "index.md")
-shutil.copyfile(PROJECT_ROOT / "img/full.png", destination / "img/full.png")
-shutil.copyfile(
-    PROJECT_ROOT / "img/docker_clients.png", destination / "img/docker_clients.png"
-)
+        module_list = pages[md_file]
```
gabrieldemarmiesse, line 96: this can be removed by using `.items()`

gabrieldemarmiesse (changes requested):
Thanks for your awesome work! The documentation looks nearly the same as before except from the code blocks and the type hints. I hope we can fix that before going forward :)

docs/autogen.py
```diff
@@ -1,147 +1,126 @@
...
 
+        for module in module_list:
+            mkdocstring_output = f"""
```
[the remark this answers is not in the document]
kclamalex: hey, good catch. I changed the overall logics once and I should got it right this time.

docs/autogen.py
kclamalex: modified.

docs/mkdocs.yml
```diff
@@ -2,8 +2,35 @@
...
+plugins:
+  - search
+  - autorefs
+  - mkdocstrings:
+      custom_templates: templates
+      default_handler: python
+      handlers:
+        python:
+          options:
+            docstring_style: sphinx
```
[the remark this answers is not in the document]
kclamalex: I have checked and I couldn't find any. However, I guess the closet we can get is google docstring style. so I changed the docstring to google style and now the type hints are there for the parameters now. You can take a look at it and see if that's what you want.

docs/mkdocs.yml
```diff
@@ -2,8 +2,35 @@
...
 site_description: 'Python on whales docs'
 
+markdown_extensions:
+- pymdownx.snippets:
+    check_paths: true
+
+plugins:
+  - search
+  - autorefs
+  - mkdocstrings:
```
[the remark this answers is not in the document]
kclamalex: couldn't find an easy way to do it as well. so the workaround i picked is to change the docstring to be google style and now the arguments are shown in table nicely. btw I need to add the material as the theme, because the table looks quite weird without theme ( word got cut into half and placed in a new line in the table ). with material theme, everything looks tidy now

gabrieldemarmiesse:
Awesome I'll take a look again, thanks you for your work!

gabrieldemarmiesse (approved):
This is awesome, it looks really good! I just pushed a fix for the requirements. I think we can merge now. Thanks for your work, the docs are even prettier than before I think :)

kclamalex:
Nice! looking forward to it!

PR #453, opened 2023-07-03
Add support for podman's `--systemd` in container run/create

### LINKED ISSUE #408: Support podman's "systemd" option when creating a container
Podman supports a `--systemd={true,false,always}` option to `podman run` or `podman create` ([docs](https://docs.podman.io/en/latest/markdown/podman-create.1.html#systemd-true-false-always)). There is no expectation that Docker will implement this (Docker actively doesn't want to support running systemd in a container), so I'm not sure how you feel about adding this option?

Unfortunately there's a Podman [bug](https://github.com/containers/podman/issues/15647) where when it tries to run a container in systemd mode it fails on non-systemd hosts using cgroups v1. The default behaviour is `--systemd=true`, which really means 'detect', i.e. run in systemd mode if the container is detected to use systemd. This means that with podman v3 a workaround is required to be able to start a systemd container on such hosts (such as Alpine 3.15).

Possible workarounds:
- Explicitly disable systemd mode (parity with Docker behaviour) with `--systemd=false`
- Manually create the `/sys/fs/cgroup/systemd/` mount on the host
- Use latest podman, use cgroups v2, or use docker...

My preferred, normal solution is the first - fundamentally this is podman adding special handling but in a broken way, so I just want to disable that special handling... Podman is much easier to get going with on Alpine, and I'm specifically wanting to test on a cgroups v1 host.

I appreciate due to this only being requested to allow working around an issue in podman and with the option not being provided by docker it might not be a very appealing thing to add to the API. However, if not all arguments will be supported it would be nice if there was at least an escape hatch to pass arbitrary arguments through to the container client rather than getting something like "TypeError: run() got an unexpected keyword argument 'systemd'".

What do you think?

LewisGaul:
Fixes #408 

- [x] Add testcase(s)

### DIFF
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -607,6 +607,7 @@ class ContainerCLI(DockerCLICaller):
         stop_timeout: Optional[int] = None,
         storage_options: List[str] = [],
         sysctl: Dict[str, str] = {},
+        systemd: Optional[Union[bool, Literal["always"]]] = None,
         tmpfs: List[ValidPath] = [],
         ulimit: List[str] = [],
         user: Optional[str] = None,
@@ -769,6 +770,7 @@ class ContainerCLI(DockerCLICaller):
 
         full_cmd.add_args_list("--storage-opt", storage_options)
         full_cmd.add_args_list("--sysctl", format_dict_for_cli(sysctl))
+        full_cmd.add_simple_arg("--systemd", systemd)
         full_cmd.add_args_list("--tmpfs", tmpfs)
         full_cmd.add_args_list("--ulimit", ulimit)
 
@@ -1318,6 +1320,7 @@ class ContainerCLI(DockerCLICaller):
         storage_options: List[str] = [],
         stream: bool = False,
         sysctl: Dict[str, str] = {},
+        systemd: Optional[Union[bool, Literal["always"]]] = None,
         tmpfs: List[ValidPath] = [],
         tty: bool = False,
         ulimit: List[str] = [],
@@ -1474,6 +1477,8 @@ class ContainerCLI(DockerCLICaller):
                 `"4g"` for example.
             stop_timeout: Signal to stop a container (default "SIGTERM")
             storage_options: Storage driver options for the container
+            systemd: Whether to run in systemd mode. Only known to apply to Podman, see
+                https://docs.podman.io/en/latest/markdown/podman-run.1.html#systemd-true-false-always
             tty: Allocate a pseudo-TTY. Allow the process to access your terminal
                 to write on it.
             user: Username or UID (format: `<name|uid>[:<group|gid>]`)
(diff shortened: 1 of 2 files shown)


LewisGaul:
@gabrieldemarmiesse I don't think there's any CI for using podman currently (see https://github.com/gabrieldemarmiesse/python-on-whales/issues/353), so I'm not sure if there's an easy way to add automated testing for this? I've tested it manually.

gabrieldemarmiesse:
That's true. Maybe let's add a unit test with a mock? That's better than nothing. You can find some other tests in the codebase that are using mocks

python_on_whales/components/container/cli_wrapper.py
```diff
@@ -1318,6 +1320,7 @@ def run(
         storage_options: List[str] = [],
         stream: bool = False,
         sysctl: Dict[str, str] = {},
+        systemd: Optional[Union[bool, Literal["always"]]] = None,
```
gabrieldemarmiesse, line 1323: can you add a test case for the boolean options? After that we can merge. I want to make sure it works without problems with `True` and `true` (i wonder if it's case sensitive)

gabrieldemarmiesse (changes requested):

python_on_whales/components/container/cli_wrapper.py
LewisGaul: The only values accepted by the API (according to the type annotation) are `None`, `True`, `False` and `"always"`, are you saying you think it should accept generic strings? See related issue: https://github.com/gabrieldemarmiesse/python-on-whales/issues/466

gabrieldemarmiesse: Ho sorry I wasn't clear, I'm not asking to change the api, I'm asking for more unit tests

gabrieldemarmiesse: Ho sorry I wasn't clear, I'm not asking to change the api, I'm asking for more unit tests

LewisGaul: > I want to make sure it works without problems with `True` and `true` (i wonder if it's case sensitive)

I'm not sure what you meant by this then?

gabrieldemarmiesse: In the unit test, we are only testing `systemd="always"`, we should also test `systemd=True` and `systemd=False`

LewisGaul: > I want to make sure it works without problems with `True` and `true` (i wonder if it's case sensitive)

Did you mean check whether podman accepts `--systemd True`, given the implementation will convert `True` to `"True"` in the command that's run? Since we can only use mock-based unit tests for podman currently and extra test won't actually check this works, but I have checked manually. I can add to the testcase anyway.

gabrieldemarmiesse: Indeed it's difficult to have podman working in the CI so mocking will be fine. I just want to make sure the function will keep working if we change the api types

LewisGaul: I've parameterised the testcase to try all three allowed values - I'm not sure it's really giving much more coverage, but let me know what you think.

gabrieldemarmiesse (approved):
Thanks! That should do the job!

PR #457, opened 2023-07-18
456 prevent to show password in exception

zimmdi:
Prevent that the password is shown in exception text


zimmdi:
Added a function to replace the value of a parameter in a blacklist with asterisk.
Added a testcase for this

zimmdi:
Better solution is to use std-in

PR #462, opened 2023-08-01
Expose `DockerClient` from the root of the package

LewisGaul:
Otherwise mypy compains and we have to do `from python_on_whales.docker_client import DockerClient`, which it feels like shouldn't be necessary?


gabrieldemarmiesse (approved):
Sure, it seems I just forgot about this one, thanks for the fix!

PR #467, opened 2023-08-11
Fix documentation for docker.compose.images()

martinburchell:
The documented workaround for lack of docker.compose.images() refers to `docker.container.ps()` instead of `docker.compose.ps()`


gabrieldemarmiesse (approved):
Thanks for the fix!

PR #473, opened 2023-09-04
fix(ContainerConfig): stop-signal can be `str` or `int`

### LINKED ISSUE #472: Container inspection fails with podman
In `components.container.models.ContainerConfig`, `stop_signal` only accepts strings. Podman otoh, responds with `StopSignal: 15`, i.e. integer values.

The line

https://github.com/gabrieldemarmiesse/python-on-whales/blob/d8851391069357b853c683b002b7bde7c4cd0234/python_on_whales/components/container/models.py#L218

should probably be `stop_signal: Optional[str | int] = None`

ambasta:
With podman, `container inspect` returns signal codes as integers, for example "StopSignal": 15

The current typing is `Optional[str]` which when enforced via pydantic, fails to parse response from podman.

Fixes #472

### DIFF
--- a/python_on_whales/components/container/models.py
+++ b/python_on_whales/components/container/models.py
@@ -215,7 +215,7 @@ class ContainerConfig(DockerCamelModel):
     mac_address: Optional[str] = None
     on_build: Optional[List[str]] = None
     labels: Optional[Dict[str, str]] = None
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse (approved):
Awesome, thanks for the fix!

Issue #475, opened 2023-09-23
Support for compse up --renew-anon-volumes


avgdev:
Hello, thanks for the great module. Will it be possible to add support for the `--renew-anon-volumes` parameter to `docker compose up`. See [up options docs](https://docs.docker.com/engine/reference/commandline/compose_up/#options)

gabrieldemarmiesse:
Sure thing! I'm open to pull requests too if you have the time :) otherwise, I'll implement it when I can

avgdev:
Sure, will submit one shortly. Thanks!

gabrieldemarmiesse:
Fixed by #487

PR #487, opened 2023-10-03
adds support for --renew-anon-volumes

avgdev:
#475 

See  https://docs.docker.com/engine/reference/commandline/compose_up/#options


gabrieldemarmiesse:
Thank you for the pull request! If you have the time, could you add a simple unit test?

Issue #485, opened 2023-10-01
Documentation deployment doesn't work anymore


gabrieldemarmiesse:
I seems since a few releases, the documentation deployment isn't working because of a jinja error. It's not a big deal because there isn't many changes between releases. It would be still very much appreciated if someone could take a look at it. You can take a look at the logs here: https://github.com/gabrieldemarmiesse/python-on-whales/actions/workflows/gh-pages.yml , also with this https://github.com/gabrieldemarmiesse/python-on-whales/blob/master/.github/workflows/gh-pages.yml it should give some info about how to reproduce locally.

Maybe a good first step is to pinpoint which commit broke the documentation?

gabrieldemarmiesse:
Fixed by #489 ! Thanks @kashyab12 !

PR #489, opened 2023-10-29
Documentation deployment doesn't work anymore

kashyab12:
#485 

**Note:** I have included fixes for each of the warnings logged during the docs' build process. According to the warning messages, type annotations wouldn't show up for the specified vars unless the warnings were fixed.

This being my first PR I worked primarily via the advice given in the description of the original bug i.e. narrow down which commit broke the docs. I tried reverting each commit one at a time (after checking out v0.64.2) however the `RecursionError` would still occur. After taking a closer look at the stack trace, and also messing with the markdown file and noticing that the presence of the `identifier` (ex: `python_on_whales.docker_client.DockerClient`) would cause the issue, I focused my efforts towards looking at issues with the `mkdocstrings-python` package. I couldn't find much about `jinja2` breaking `mkdocs` (except for https://github.com/mkdocs/mkdocs/issues/2799), however I did find good amounts of discussion regarding `Griffe` causing issues (https://github.com/mkdocstrings/mkdocstrings/issues/382, and https://github.com/mkdocstrings/griffe/issues/79). According to the [changelog](https://github.com/mkdocstrings/python/blob/main/CHANGELOG.md), `mkdocstrings-python` enforced a version upper bound on `griffe` in version `1.3`, hence why I chose to bump to this particular version. I tried out 1.2 as well however still ran into the same issue.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,5 @@
 mkdocstrings==0.20.0
-mkdocstrings-python==1.0.0
+mkdocstrings-python==1.3.0
 mkdocs-autorefs==0.4.1
 mkdocs-material==9.1.14
 pymdown-extensions==10.0.1
--- a/python_on_whales/components/compose/cli_wrapper.py
+++ b/python_on_whales/components/compose/cli_wrapper.py
@@ -37,7 +37,7 @@ class ComposeCLI(DockerCLICaller):
             services: The services to build (as list of strings).
                 If `None` (default), all services are built.
                 An empty list means that nothing will be built.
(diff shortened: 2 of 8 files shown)


gabrieldemarmiesse:
Thank you for this thorough investigation. I'll take a look at this PR this week

gabrieldemarmiesse:
I just made a copy of your branch and ran the deployment job for the docs from this, and it worked like a charm. Thank you for the version fix and all the other small changes to the docstrings. This is really the kind of pull requests that I love, they just work :)  This PR makes me glad I'm doing open source. Many thanks :)

Issue #490, opened 2023-10-31
Fields of type"typing.Annotated" are not supported in pydantic


maarten0912:
When importing `python_on_whales` with `pydantic` version 1.7.4, I get the following error:

```
    from python_on_whales import docker
  File "/lib/python3.10/site-packages/python_on_whales/__init__.py", line 2, in <module>
    from .components.buildx.cli_wrapper import Builder
  File "/lib/python3.10/site-packages/python_on_whales/components/buildx/cli_wrapper.py", line 9, in <module>
    import python_on_whales.components.image.cli_wrapper
  File "/lib/python3.10/site-packages/python_on_whales/components/image/cli_wrapper.py", line 16, in <module>
    from python_on_whales.components.container.cli_wrapper import (
  File "/lib/python3.10/site-packages/python_on_whales/components/container/cli_wrapper.py", line 30, in <module>
    from python_on_whales.components.container.models import (
  File "/lib/python3.10/site-packages/python_on_whales/components/container/models.py", line 55, in <module>
    class ContainerDeviceRequest(DockerCamelModel):
  File "/lib/python3.10/site-packages/pydantic/main.py", line 262, in __new__
    fields[ann_name] = inferred = ModelField.infer(
  File "/lib/python3.10/site-packages/pydantic/fields.py", line 315, in infer
    return cls(
  File "/lib/python3.10/site-packages/pydantic/fields.py", line 284, in __init__
    self.prepare()
  File "/lib/python3.10/site-packages/pydantic/fields.py", line 356, in prepare
    self._type_analysis()
  File "/lib/python3.10/site-packages/pydantic/fields.py", line 510, in _type_analysis
    raise TypeError(f'Fields of type "{origin}" are not supported.')
TypeError: Fields of type "<class 'typing.Annotated'>" are not supported.
```

Maybe a newer version of pydantic is required, while `requirements.txt` only requires version 1.5 or higher, except 2.0.x.

Reference for [Annotated](https://docs.pydantic.dev/latest/concepts/types/#custom-types)

gabrieldemarmiesse:
Thank you for the bug report! Indeed it seems that the version range is wrong for pydantic. We should find which version of pydantic allows Annotated and us that version as the lower bound

gabrieldemarmiesse:
This is fixed by @kashyab12  in #491 , I'll make a new release.

PR #491, opened 2023-11-03
Fields of type"typing.Annotated" are not supported in pydantic

kashyab12:
#490 

Reproduced the issue with `pydantic==1.7.4`. After looking at the `pydantic` [changelog](https://docs.pydantic.dev/latest/changelog/#v18-2021-02-26) `v1.8 2021-02-26` introduced support for `typing.Annotated` hints on model fields. After installing `pydantic==1.8` and trying to reproduce the issue, no errors were produced.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -62,7 +62,7 @@ jobs:
           - test_task.py
           - test_volume.py
         pydantic-version:
-          - pydantic==1.*
+          - pydantic==1.8
           - pydantic==2.*
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -62,7 +62,7 @@ jobs:
           - test_task.py
           - test_volume.py
         pydantic-version:
-          - pydantic==1.8
+          - pydantic==1.9
           - pydantic==2.*
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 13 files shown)


gabrieldemarmiesse:
Thanks! In the github yaml files, the pydantic version is hardcoded for the unit tests, can you set it to 1.8.0 instead of just 1.* so that our unit tests run with the oldest version possible? Thanks!

kashyab12:
> Thanks! In the github yaml files, the pydantic version is hardcoded for the unit tests, can you set it to 1.8.0 instead of just 1.* so that our unit tests run with the oldest version possible? Thanks!

@gabrieldemarmiesse after modifying the `pydantic` version in the github jobs, lots of tests are failing. Upon taking a look at the stacktrace seems like the `pydantic` object is missing the `__version__` attribute. [This issue](https://github.com/pydantic/pydantic/issues/2572) was fixed in [version 1.9 according to the changelog](https://docs.pydantic.dev/latest/changelog/#v190a1-2021-12-18-changes). So either:
- We modify the job to use version 1.9 (not the oldest version), which wouldn't tell us whether a new change breaks compatability with version 1.8; or
- Modify both minimum supported version and job version to use 1.9; or
- Modify references to `__version__` and obtain the version by some other means.

gabrieldemarmiesse:
We can se the minimul version to 1.9. That why we have our CI!

kashyab12:
@gabrieldemarmiesse I've updated CI to use `pydantic==1.9`.

gabrieldemarmiesse (approved):

gabrieldemarmiesse:
Many thanks!

PR #492, opened 2023-11-07
Add platform variant to image inspect

d4nj1:
Hi,
I tried to look up the variant property on images where this value is set and it failed. When looking into it I saw that it was missing in the library. So here you get a PR. :wink: I also added a test case to verify the change.

Thanks for your efforts on this project. It's truly awesome!


gabrieldemarmiesse (approved):

gabrieldemarmiesse:
This awesome project is largely done by contributors (I think the ratio of contributors/users is quite high judging by the numbers haha), so thank you for your efforts on this project :)

PR #497, opened 2023-11-11
Allow the streaming of the docker.compose.pull logs

HemaZ:
same as #496 but for compose.pull. The test will most probably fail but i couldn't run it locally. having some problem with the docker registry, so hopefully it will run on github action and i will update it to the expected output

### DIFF
--- a/python_on_whales/components/compose/cli_wrapper.py
+++ b/python_on_whales/components/compose/cli_wrapper.py
@@ -434,13 +434,36 @@ class ComposeCLI(DockerCLICaller):
             for proj in json.loads(run(full_cmd))
         ]
 
+    @overload
+    def pull(
+        self,
+        services: Union[List[str], str, None] = ...,
+        ignore_pull_failures: bool = ...,
+        include_deps: bool = ...,
+        quiet: bool = ...,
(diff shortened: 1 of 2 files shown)


HemaZ:
oh so it seems the docker hub issue is a problem with docker hub itself.

gabrieldemarmiesse:
Indeed I had the same issue. Dockerhub seems to be having a bad time

HemaZ:
> Indeed I had the same issue. Dockerhub seems to be having a bad time

I think it got fixed. This passes now

gabrieldemarmiesse (approved):
Many thanks for the pull request! You're the 50th contributor of Python-on-whales! Sadly I don't have any prize, but you can be proud to add another cool feature to the lib that many users will appreciate!

HemaZ:
> Many thanks for the pull request! You're the 50th contributor of Python-on-whales! Sadly I don't have any prize, but you can be proud to add another cool feature to the lib that many users will appreciate!

Thank you for creating this amazing project. It saved me a lot of time and i really appreciate it.

gabrieldemarmiesse:
I'm happy to hear it! The nice words of people like you are a good reason for me to work in this library :)

Issue #466, opened 2023-08-11
`docker.legacy_build(pull="never")` is treated as `pull=True`


LewisGaul:
It seems that APIs that only accept booleans will be interpreted as a boolean value even if another type is given. This is particularly confusing in the case of `docker.legacy_build(pull: bool)`, which is subtly different to `docker.run(pull: str)`. Would it be worth doing some arg type validation, at least in the `Command.add_flag()` method?

gabrieldemarmiesse:
Yes I think that would be helpful. I would be happy to get such a PR :)

LewisGaul:
> Maybe we should consider using [beartype](https://beartype.readthedocs.io/en/latest/) for this. That would avoid some complexity in our arguments. Maybe we should decorate our public api with this?

That sounds like an interesting idea, I hadn't come across beartype before - what enigmatic docs!

kashyab12:
> EDIT: Maybe we should consider using [beartype](https://beartype.readthedocs.io/en/latest/) for this. That would avoid some complexity in our arguments. Maybe we should decorate our public api with this?

@gabrieldemarmiesse would the following suffice?
```
from beartype import beartype
...
    @beartype
    def add_flag(self, name: str, value: bool):
        if value:
            self.append(name)
```
I asserted that `beartype` catches these issues at runtime, so this should solve `str` types being passed. Moreover, the `if value` check can also removed as a result of this.

gabrieldemarmiesse:
The issue with adding the beartype decorator used in non-public APIs is that the error messages won't be clean for the users. The error message will talk about a type mismatch in an internal function and it won't be obvious what caused this. I tend to think that decorating public apis functions with beartype would make error messages clearer from a user perspective.

kashyab12:
That makes lots of sense, thanks for the explanation! I'm going to read more `beartype` docs before trying to work a fix. Appreciate the feedback!

Issue #463, opened 2023-08-03
Error docker.execute with str


sylvainmouquet:
Hello,

I have an error when i try to run: 
`docker.execute(vault_server.id, command="vault login -address=http://localhost:8200 myroot")`

> TypeError: When calling docker.execute(), the second argument ('command') should be a list. Here are some examples:docker.execute('somecontainer', ['ls']), docker.execute('somecontainer', ['cat', '/some/file.txt']) In your case, command should not be a string. You can try docker.execute('test-vault-b5d0d7ec-b13d-447f-a60f-185e68555dc4', ['vault', 'login', '-address=http://localhost:8200', 'myroot'], ...).

but in the definition of the execute command, the str should be authorized : 
```

    def execute(
        self,
        container: ValidContainer,
        command: Union[str, List[str]],

```

gabrieldemarmiesse:
Thank you for this issue. Indeed, the type hint do not correspond to what python-on-whales expect. That's a mistake on my part.

We cannot use sh when the user provides a str instead of a list[str] because sh is not available in all containers.

We should change the type hints `Union[str, list[str]]` -> `list[str]` to be clear about what the method expects

gabrieldemarmiesse:
Closing as fixed.

PR #501, opened 2023-11-14
Error docker.execute with str

kashyab12:
#463 

Getting rid of `str` annotations from `command` variables.  I did notice that the `execute` method for the `ContainerCLI` class explicitly asserts for the `command` var to be of `list` type and throws an appropriate exception message if it is not. I decided not to perform the same assertion within the `ComposeCLI` class's `execute` function and the `ServiceCLI` class's `create` function since `beartype` can be of value of here (see comments within #466) . However, I am open to feedback regarding this.

### DIFF
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -261,7 +261,7 @@ class Container(ReloadableObjectFromJson):
 
     def execute(
         self,
-        command: Union[str, List[str]],
+        command: List[str],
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse:
Thank you for the pull request. Indeed, I'm still thinking about beartype from time to time, weighting the pros and cons (we might loose duck typing unless we use more complex types, like protocols).

Anyways, your PR is a strict improvement compared to the previous behaviour, so let's merge it

Issue #499, opened 2023-11-12
Add stream_logs argument to all compose functions that can stream some output


gabrieldemarmiesse:
Every compose fonction that logs something interesting (up, exec, run, push, pull, start, build) should have a new argument (or rename the existing one): stream_logs. When this is enabled, an iterator is returned. This iterator is a tuple[str, bytes]. You can see how this iterator looks like here: https://gabrieldemarmiesse.github.io/python-on-whales/user_guide/docker_run/#stream-the-output

If the user set in any function stream_logs=True and quiet=True at the same time, Python-on-whales should throw an error saying that it's not possible to have both arguments being True at once.

When stream_logs=False, then either None is returned, or the already existing object, like in the function docker.compose.run() which already does that and returns a Container when not streaming.

If a function already has a stream argument, then this argument must be changed to stream_logs.

All of that is possible with the function stream_stdout_and_stderr. An example can be found with docker.compose.run().

To help with the conditional type return, we should use the typing.overload decorator. It will allow mypy, pyright, vscode and pycharm tell the user when the iterator is returned or when something else is returned depending on the input arguments. So the conditional type on return should not be a big deal.

Here are the functions to modify

[x] `docker.compose.build` #494
[x] `docker.compose.create` #498
[x] `docker.compose.down` #496
[] `docker.compose.execute`
[] `docker.compose.kill`
[] `docker.compose.pause`
[x] `docker.compose.pull` #497
[] `docker.compose.push`
[] `docker.compose.restart`
[] `docker.compose.rm`
[] `docker.compose.start`
[] `docker.compose.stop`
[] `docker.compose.unpause`
[] `docker.compose.up`

If you're interested in this ticket, please make one pull request per function. And if you're looking for examples, you can take a look at the pull requests already done on this subject.

PR #505, opened 2023-11-19
✨ Added stream_logs to docker.compose.up()

HemaZ:
part of #499 . adding stream_logs to the compose.up() function.


gabrieldemarmiesse (approved):
Many thanks!

PR #507, opened 2023-11-25
✨ Added stream_logs to docker.compose.stop()

HemaZ:
part of https://github.com/gabrieldemarmiesse/python-on-whales/issues/499 . adding stream_logs to the compose.stop() function.


gabrieldemarmiesse (approved):
Many thanks!!

PR #510, opened 2023-11-28
adding support for docker stack config

### LINKED ISSUE #509: add support for docker stack config command
Hello,

I notice that the StackCLI class miss the `config` method (see https://docs.docker.com/engine/reference/commandline/stack_config/).

I'll create a pull request :-)

thosil:
The `docker stack config` command returns a yaml content on stdout. We have to parse the yaml output, hence the new pyyaml requirement.

The parameter "return_json" like the `ComposeCLI.config` method.

+ a couple of tests

fix: #509

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ requests
 tqdm
 typer>=0.4.1
 typing_extensions
+pyyaml >= 6.0.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,3 @@ requests
 tqdm
 typer>=0.4.1
 typing_extensions
-pyyaml >= 6.0.1
--- a/.gitignore
+++ b/.gitignore
@@ -129,3 +129,6 @@ dmypy.json
 .pyre/
 docs/generated_sources
 docs/site
+
+# pycharm or other jetbrains ide
+.idea/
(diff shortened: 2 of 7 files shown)


requirements.txt
```diff
@@ -3,3 +3,4 @@ requests
 tqdm
 typer>=0.4.1
 typing_extensions
+pyyaml >= 6.0.1
```
gabrieldemarmiesse, line 6: I think this should be an optional dependency as only a small percentage of users will actually use `docker.stack.config()`. We don't want to force this dependency on everyone.

gabrieldemarmiesse (changes requested):

requirements.txt
thosil: ok, pyyaml is optional in the latest commit and could be installed as extra (pip install python_on_whales[yaml])
I also had to try importing the yaml module and handle the fact it could be missing in the config() method.

I don't know if it's the better way to do it.

thosil:
:grimacing: I forgot to add a docstring and I don't have the time to do it for the moment, sorry

gabrieldemarmiesse:
No worries, we're not in a hurry. In the open source world, we have all the time we want

gabrieldemarmiesse (changes requested):

gabrieldemarmiesse:
Closing as stale. Feel free to open another PR if you want!

PR #511, opened 2023-11-29
service create args: network,restart,secrets

fuentes73:
These arguments are useful for creating "one-shot" services. Like db-migrations, backups, etc...

### DIFF
--- a/python_on_whales/components/service/cli_wrapper.py
+++ b/python_on_whales/components/service/cli_wrapper.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from datetime import datetime, timedelta
-from typing import Any, Dict, List, Optional, Union, overload
+from typing import Any, Dict, List, Literal, Optional, Union, overload
 
 import python_on_whales.components.task.cli_wrapper
 from python_on_whales.client_config import (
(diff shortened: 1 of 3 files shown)


python_on_whales/components/service/cli_wrapper.py
```diff
@@ -173,6 +173,10 @@ def create(
         limit_memory: Optional[str] = None,
         limit_pids: Optional[int] = None,
         log_driver: Optional[str] = None,
+        network: Optional[str] = None,
+        restart_condition: Optional[str] = None,
```
gabrieldemarmiesse, line 177: I'm thinking about the type hint for restart_condition. It could be `Optional[Literal["none", "on-failure", "any"]]` no? What do you think?

python_on_whales/components/service/cli_wrapper.py
```diff
@@ -173,6 +173,10 @@ def create(
...
+        restart_max_attempts: Optional[int] = None,
+        secrets: List[str] = [],
```
gabrieldemarmiesse, line 179: For secret, reading the doc https://docs.docker.com/engine/reference/commandline/service_create/#secret I think it could be `Union[dict[str, str], str, None]` and we could craft the string `source=app-key,target=app,uid=1000,gid=1001,mode=0400` from a dict, that would avoid some string manipulation by the user. What do you think?

fuentes73: Done! The changes are not exactly as you had proposed but let me know what do you think.

Issue #512, opened 2023-12-11
Make typer/tqdm/requests optional dependencies?


DanCardin:
As a user of python-on-whales in its capacity as a library (thanks for the library by the way!) I noticed the dependency on typer/tqdm and thought it looked weird.

After looking briefly, it seems to me like at least typer could straightforwardly be shifted to a `cli` extra and definitely be completely optional.

tqdm seems to be used a bit more deeply, but would also seem to me to only be a relevant dependency for someone using the CLI. With a bit more work, I could imagine it too being made optional.

Would you accept a PR doing so?

gabrieldemarmiesse:
I would be ok to make typer optional, but for tqdm, it may be needed if the user doesn't have the CLI locally because it will be downloaded automatically. So I'm afraid it can't be optional :(

DanCardin:
I realize that the code that utilizes tqdm is invoked during normal library operations also. but from the perspective of someone importing this library (and running it rather than invoking the CLI), having a tqdm progress bar in my logs is kind of actively the opposite of what I'd want. Although honestly, the automatic downloading of the docker cli itself, is even functionality that i'll never encounter the environments in which I'm using python-on-whales.

> Out of curiosity, why are you concerned about the number of dependencies?

It's perhaps partly a purity thing. I dont want to have to install typer/tqdm when I **know** that they'll never be used and the functionality they provide **is** completely optional.

More generally, I do think it has positive downstream effects. You may (or may not) be judicious about the versions you depend upon, but adding typer as a dependency just increases the possibility that I or one of my dependencies might depend on a different version that you and cause a conflict (again, for a piece of functionality I wont use).

gabrieldemarmiesse:
if you don't want to have a progress bar in the logs, you just have to make sure the docker CLI is installed before running python-on-whales. The auto-download is just here for quick bootstrapping on a new system.

I would like to be able to vendor tqdm, but sadly, vendoring is not supported well in python, so it's not something I can do. I'll accept a PR to make typer optional though.

LewisGaul:
For what it's worth, I'm in favour of reducing dependencies, where making them optional is a reasonable approach as long as it doesn't diminish the user experience. The problem is doing that without breaking backwards compatibility, as raised in the linked PR.

[...]

PR #513, opened 2023-12-11
refactor: Make typer an optional dependency.

DanCardin:
Fixes https://github.com/gabrieldemarmiesse/python-on-whales/issues/512

### DIFF
--- a/python_on_whales/command_line_entrypoint.py
+++ b/python_on_whales/command_line_entrypoint.py
@@ -1,6 +1,11 @@
 from typing import Optional
 
-import typer
+try:
+    import typer
+except ImportError as exc:
+    raise ImportError(
+        "The `python-on-whales` CLI, requires the 'cli' extra. `pip install python-on-whales[cli]`."
+    ) from exc
 
 from python_on_whales import docker
 from python_on_whales.download_binaries import download_docker_cli
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -22,7 +22,7 @@ jobs:
         sudo service docker restart
         sleep 2
         docker info
-        pip install -e .
+        pip install -e '.[cli]'
         pip install -r ./lint-requirements.txt
         flake8
         isort --check ./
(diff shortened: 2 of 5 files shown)


DanCardin:
Ideally I would add to this that `tqdm` is also part of the "cli" extra. perhaps also add a progress=False flag (which defaults to False if tqdm were not installed), but the CLI defaults to `True`. In my opinion, the vanilla calling of `python_on_whales` through code shouldn't be enabling tqdm progress bars by default (except through the CLI). As a library, this sort of output is very surprising.

However, I didn't include that so far given your comments in the original issue.

gabrieldemarmiesse:
Thank you for the pull request. I didn't expect it to be that fast! I think that first we need to take into account backward compatibility. By merging this right now, everyone relying on the CLI will have their CI pipeline broken. We should warn them first, for a few releases, then set typer as an optional dependency. Suddenly adding a breaking change to the software is not something that we should do unless there is a security issue.

Let's leave this PR open, we can merge it in a few releases. If you're willing to add the warning, I would happily merge such a PR now.

DanCardin:
The problem, as I see it, with warning; is that I dont see that there's anything the user can do to make the warning go away. 

I could add the extra, but leave the dependency as required. But as far as I know, there's no way to detect if a library was installed with an extra or not, so they'll get the warning even if they have switched to using the cli extra.

I would think use of the CLI would primarily be a local-use tool, rather than something that would break someone's build. but that's completely hearsay from my own personal use.

python_on_whales/command_line_entrypoint.py
```diff
@@ -1,6 +1,11 @@
 from typing import Optional
 
-import typer
+try:
+    import typer
+except ImportError:
+    raise ImportError(
+        "The `python-on-whales` CLI, requires the 'cli' extra. `pip install python-on-whales[cli]`."
+    )
```
LewisGaul, line 8: There should be a `from` here, either `from None` or `from exc` (making it `except ImportError as exc`).

gabrieldemarmiesse:
I think this can be closed as typer is not a dependency anymore! 🎉

PR #514, opened 2023-12-13
feat(service): Add --mount arg && update --replicas

fuentes73:
- Add the posibility of mount volumes when creating a service
- Add the posiblity of specify `--replicas` & `--quiet` when updating a service

### DIFF
--- a/python_on_whales/components/service/cli_wrapper.py
+++ b/python_on_whales/components/service/cli_wrapper.py
@@ -120,6 +120,8 @@ class Service(ReloadableObjectFromJson):
         force: bool = False,
         image: Optional[str] = None,
         with_registry_authentication: bool = False,
+        quiet: bool = False,
+        replicas: Optional[int] = None,
     ):
         """Updates a service
 
@@ -127,7 +129,7 @@ class Service(ReloadableObjectFromJson):
         information about the arguments.
         """
         ServiceCLI(self.client_config).update(
-            self, detach, force, image, with_registry_authentication
+            self, detach, force, image, with_registry_authentication, quiet, replicas
         )
 
     def exists(self) -> bool:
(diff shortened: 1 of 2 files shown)


python_on_whales/components/service/cli_wrapper.py
```diff
@@ -119,6 +119,8 @@ def update(
         detach: bool = False,
         force: bool = False,
         image: Optional[str] = None,
+        quiet: bool = False,
+        replicas: Optional[int] = None,
```
gabrieldemarmiesse, line 123: Placing them here would break backward compatibility in a silent manner for people who use positional arguments. Could you add them at the end, after the argument `with_registry_authentication`? Many thanks!

python_on_whales/components/service/cli_wrapper.py
```diff
@@ -173,6 +175,7 @@ def create(
         limit_memory: Optional[str] = None,
         limit_pids: Optional[int] = None,
         log_driver: Optional[str] = None,
+        mounts: Optional[List[Dict[str, str]]] = [],
```
gabrieldemarmiesse, line 178: Same comment here, they should be at the end.

python_on_whales/components/service/cli_wrapper.py
```diff
@@ -463,6 +469,8 @@ def update(
         detach: bool = False,
         force: bool = False,
         image: Optional[str] = None,
+        quiet: bool = False,
+        replicas: Optional[int] = None,
```
gabrieldemarmiesse, line 473: Same here, we should add it after `with_registry_authentication`

gabrieldemarmiesse (changes requested):
Very nice pull request, many thanks, we need to touch a few things to ensure backward compatibility.

python_on_whales/components/service/cli_wrapper.py
fuentes73: You're right, fixed!

gabrieldemarmiesse (approved):

gabrieldemarmiesse:
Thanks again for the feature :)

PR #517, opened 2024-01-01
Tidy up test/SA CI run

LewisGaul:
- Split up CI setup script into separate responsibilities
- Tidy up use of "setup" phase for each run
- Comment out the no-binary run, since it wasn't actually running without binaries?
- Don't fail fast on the matrix run, so that all failures can be checked at once
- Use 'master' rather than fixed version for checkout and setup-python actions to avoid node warnings (old versions seem more likely to break than being on latest)
- Don't bother setting up docker for linting

Precursor to adding podman to the CI.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -10,8 +10,8 @@ jobs:
   build-linux-lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
-      - uses: actions/setup-python@master
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: '3.8'
       - name: Setup
@@ -51,8 +51,8 @@ jobs:
           - pydantic==2.*
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
-      - uses: actions/setup-python@master
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: '3.8'
       - name: Setup
@@ -66,8 +66,8 @@ jobs:
   build-linux-test-other:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
-      - uses: actions/setup-python@master
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: '3.8'
       - name: Setup
@@ -80,8 +80,8 @@ jobs:
   build-windows:
     runs-on: windows-latest
     steps:
-      - uses: actions/checkout@master
-      - uses: actions/setup-python@master
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: 3.8
       - name: Setup
@@ -99,8 +99,8 @@ jobs:
 #  build-macos:
 #    runs-on: macos-latest
 #    steps:
-#      - uses: actions/checkout@master
-#      - uses: actions/setup-python@master
+#      - uses: actions/checkout@v4
+#      - uses: actions/setup-python@v5
 #        with:
 #          python-version: 3.8
 #      - name: Setup
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -1,6 +1,5 @@
 name: Run tests
 
-
 on:
   push:
     branches:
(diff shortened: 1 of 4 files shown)


LewisGaul:
@gabrieldemarmiesse I think the list of required tests will need updating, which I don't seem to have permissions to do. I'm looking into the strange import issue...

LewisGaul:
Looks like the only issue now is:
```
ERROR tests/python_on_whales/components/test_manifest.py::test_manifest_platform_variant - python_on_whales.exceptions.DockerException: The docker command executed was `/usr/bin/docker image pull --quiet arm64v8/busybox:1.35`.
It returned with code 1
The content of stdout is ''
The content of stderr is 'no matching manifest for linux/amd64 in the manifest list entries
```
I'm not sure what's causing this, and don't see why it would be related to these changes...

gabrieldemarmiesse:
Maybe let's try to split this PR into multiple ones? It will be easier to understand why the CI is unhappy

LewisGaul:
It looks like the failure is seen on master branch, and is not related to my changes (which I think make sense as a single PR). Here's a branch which simply gets the CI to run (https://github.com/gabrieldemarmiesse/python-on-whales/compare/master...LewisGaul:python-on-whales:CI-test), and here's the same failure seen: https://github.com/LewisGaul/python-on-whales/actions/runs/7380593195/job/20078127537

LewisGaul:
Raised https://github.com/gabrieldemarmiesse/python-on-whales/issues/518 to track the test breakage.

LewisGaul:
Cool, the CI is all passing now, but still need the required tests updating since removing/renaming some (that's why it shows as yellow).

.github/workflows/python-package.yml
```diff
@@ -11,38 +10,24 @@ jobs:
...
-    - uses: actions/setup-python@v2
-      with:
-        python-version: '3.8'
-    - name: Run all tests
-      run: |
-        bash tests/ci-setup.sh
-        python -m pytest -v --durations=10 --ignore=tests/python_on_whales/components
-
+      - uses: actions/checkout@master
+      - uses: actions/setup-python@master
```
gabrieldemarmiesse, line 14: You mentionned it was better to use master, but it seems an anti-pattern to unpin versions of software we're using in the CI, as a new version can break our CI without any action from our side. Can you expand a bit on the motivation?

.github/workflows/python-package.yml
```diff
@@ -66,71 +51,89 @@ jobs:
...
+        with:
+          python-version: '3.8'
+      - name: Setup
+        run: |
+          ./scripts/ci-setup.sh
+      - name: Run tests
+        run: |
+          python -m pytest -vv --durations=10 --ignore=tests/python_on_whales/components/
+
+#  build-linux-test-without-any-binary:
```
gabrieldemarmiesse, line 80: I think you can remove this section.

gabrieldemarmiesse (commented):
Very nice cleaning, I'll fix the settings for the tests. Can you expand a bit on the unpinning of our actions?

.github/workflows/python-package.yml
LewisGaul: Ok, will do :)

.github/workflows/python-package.yml
LewisGaul: I'm actually not set on this change - happy to change to newer versions that don't cause the warnings seen if you scroll down on a run, e.g. [here](https://github.com/gabrieldemarmiesse/python-on-whales/actions/runs/7411813041).

The goal is to avoid the CI breaking due to an external change, as you stated. There are two options with separate reasons it could break:
- Don't pin (use latest/master) and a new version could have a breaking change
- Do pin, and GH actions could drop support for something it depends on

We're seeing warnings because of being on old versions and GH actions no longer supports the version of node the action was written to work with - this could in theory cause the action to break. On the other hand, we use hardly any customisation options with the actions, which are highly unlikely to break in future versions. So it seemed to me perhaps the balance is more in favour of just using latest, but in reality it probably doesn't matter and I was mainly just seeking to get rid of the warnings!

gabrieldemarmiesse: In a perfect world, we would pin the version of every dependency/software we use in the CI, and a bot would open pull requests to bump the versions when a new one is available, ensuring we are always up to date and don't use outdated stuff that github might remove. 
This would also ensure that we have perfect reproducibility of our builds, which is always nice. But yeah, we depend on github actions which is a moving target in terms of pre-installed software.

Even if it's not fool-proof, I'd prefer that we stick to best practices and pin the versions. I agree that we might get warnings if we don't change the version for a while, but at least, if on the action side of things, they push a bug in their master branch, our CI won't break, which is a good thing.

long story short, I'd prefer that we just upgrade the versions and do not use master.

gabrieldemarmiesse (approved):
Thanks!

Issue #518, opened 2024-01-02
Broken test: `test_manifest.py::test_manifest_platform_variant`


LewisGaul:
Test fails in CI with:
```
ERROR tests/python_on_whales/components/test_manifest.py::test_manifest_platform_variant - python_on_whales.exceptions.DockerException: The docker command executed was `/usr/bin/docker image pull --quiet arm64v8/busybox:1.35`.
It returned with code 1
The content of stdout is ''
The content of stderr is 'no matching manifest for linux/amd64 in the manifest list entries
```

Also seen running manually with docker and podman:
```
$docker pull arm64v8/busybox:1.35
1.35: Pulling from arm64v8/busybox
no matching manifest for linux/amd64 in the manifest list entries

$podman pull arm64v8/busybox:1.35
✔ docker.io/arm64v8/busybox:1.35
Trying to pull docker.io/arm64v8/busybox:1.35...
Error: choosing an image from manifest list docker://arm64v8/busybox:1.35: no image found in image index for architecture amd64, variant "", OS linux
```

LewisGaul:
This testcase was added only a couple of months ago in https://github.com/gabrieldemarmiesse/python-on-whales/pull/492, although it passed in the CI at the time. I'm not sure exactly what's changed here to cause the breakage though, perhaps the docker API/registry no longer supports pulling images that are for an architecture not supported by the host?

Note two of the other testcases in that file were also marked to be skipped in CI back in May...

d4nj1:
Hi @LewisGaul ,
could it be that the CI system changed from arm64 to amd64? At least the error message complains about platform not matching. There are two things you could test:

1. Force downloading the arm64 busybox image anyways with`docker pull --platform linux/arm64 arm64v8/busybox:1.35`
2. Switch to amd64 by changing the pull command to `docker pull busybox:1.35`

Does this help?

d4nj1:
Hi @LewisGaul ,
I looked a bit deeper and I think two thinks come together here.

1. Docker pull refuses to pull images for other platforms if they are not specified with `--platform=PLATFORM`.
Output of `docker pull --help`

```
Usage:  docker pull [OPTIONS] NAME[:TAG|@DIGEST]

Download an image from a registry

Aliases:
  docker image pull, docker pull

Options:
  -a, --all-tags                Download all tagged images in the repository
      --disable-content-trust   Skip image verification (default true)
      --platform string         Set platform if server is multi-platform capable
[...]

PR #519, opened 2024-01-03
#518 Add platform option for pull

d4nj1:
This also fixes the platform variant test case mentioned in  #518 

**docker pull --help**

```
Usage:  docker pull [OPTIONS] NAME[:TAG|@DIGEST]

Download an image from a registry

Aliases:
  docker image pull, docker pull

Options:
  -a, --all-tags                Download all tagged images in the repository
      --disable-content-trust   Skip image verification (default true)
      --platform string         Set platform if server is multi-platform capable
  -q, --quiet                   Suppress verbose output

```

### DIFF
--- a/python_on_whales/components/image/cli_wrapper.py
+++ b/python_on_whales/components/image/cli_wrapper.py
@@ -476,7 +476,10 @@ class ImageCLI(DockerCLICaller):
         return run(full_cmd)
 
     def pull(
-        self, x: Union[str, List[str]], quiet: bool = False
+        self,
+        x: Union[str, List[str]],
+        quiet: bool = False,
+        platform: Optional[str] = None,
     ) -> Union[Image, List[Image]]:
         """Pull one or more docker image(s)
 
@@ -488,6 +491,7 @@ class ImageCLI(DockerCLICaller):
                 The progress bars might look strange as multiple
                 processes are drawing on the terminal at the same time.
             quiet: If you don't want to see the progress bars.
+            platform: If you want to enforce a platform.
 
         Returns:
             The Docker image loaded (`python_on_whales.Image` object).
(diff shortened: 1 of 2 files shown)


python_on_whales/components/image/cli_wrapper.py
```diff
@@ -476,7 +476,7 @@ def prune(self, all: bool = False, filter: Dict[str, str] = {}) -> str:
         return run(full_cmd)
 
     def pull(
-        self, x: Union[str, List[str]], quiet: bool = False
+        self, x: Union[str, List[str]], quiet: bool = False, platform: str = ""
```
LewisGaul, line 479: Would be better to make this `platform: Optional[str] = None` to be clear about the intent of not specifying a platform (rather than specifying an empty string potentially having meaning).

python_on_whales/components/image/cli_wrapper.py
```diff
@@ -498,23 +502,26 @@ def pull(
...
             pool = ThreadPool(4)
-            generator = self._generate_args_push_pull(x, quiet)
+            generator = self._generate_args_pull(x, quiet, platform)
             all_images = pool.starmap(self._pull_single_tag, generator)
             pool.close()
             pool.join()
             return all_images
 
-    def _pull_single_tag(self, image_name: str, quiet: bool):
+    def _pull_single_tag(self, image_name: str, quiet: bool, platform: str):
```
LewisGaul, line 516: Sorry I think these function signatures (also `_generate_args_pull()`) need to be changed to `platform: Optional[str]` since a `None` may be passed through now.

d4nj1:
I only squashed the commits

gabrieldemarmiesse (approved):
Very nice, this fix is much appreciated. In the future, we might think of allowing the `platform` argument to be a sequence.

PR #520, opened 2024-01-06
Support test parameterisation by docker and podman, enable in CI

LewisGaul:
- `pytest tests/` - run tests with docker, also with podman if available
- `pytest tests/ --ctr-exe docker` - run tests with docker only
- `pytest tests/ --ctr-exe podman` - run tests with podman only
- `pytest tests/ --ctr-exe docker,podman` - explicitly run tests with both docker and podman (for use in the CI)
- Also supports non-standard exe names and detects whether docker/podman, e.g. `pytest tests/ --ctr-exe podman4.2`

With the default, which checks whether docker and podman are available before parameterising on both:
```
(venv) python-on-whales/(podman-CI)$python -m pytest tests/python_on_whales/components/test_image.py -k test_filter_when_listing_old_signature -vv
=============================================== test session starts ================================================
platform linux -- Python 3.9.5, pytest-7.1.2, pluggy-1.0.0 -- /mnt/c/Users/legaul/repos/python-on-whales/venv/bin/python
cachedir: .pytest_cache
rootdir: /mnt/c/Users/legaul/repos/python-on-whales
plugins: mock-3.8.2
collected 66 items / 64 deselected / 2 selected

tests/python_on_whales/components/test_image.py::test_filter_when_listing_old_signature[docker] PASSED       [ 50%]
tests/python_on_whales/components/test_image.py::test_filter_when_listing_old_signature[podman] SKIPPED (Not supported with podman) [100%]

=================================== 1 passed, 1 skipped, 64 deselected in 1.46s ====================================
```

Specifying only docker:
```
(venv) python-on-whales/(podman-CI)$python -m pytest tests/python_on_whales/components/test_image.py -k test_filter_when_listing_old_signature -vv --ctr-exe docker
=============================================== test session starts ================================================
platform linux -- Python 3.9.5, pytest-7.1.2, pluggy-1.0.0 -- /mnt/c/Users/legaul/repos/python-on-whales/venv/bin/python
cachedir: .pytest_cache
rootdir: /mnt/c/Users/legaul/repos/python-on-whales
plugins: mock-3.8.2
collected 40 items / 39 deselected / 1 selected

tests/python_on_whales/components/test_image.py::test_filter_when_listing_old_signature[docker] PASSED       [100%]

========================================= 1 passed, 39 deselected in 1.54s =========================================
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,7 +3,7 @@ import subprocess
 import tempfile
 import time
 from pathlib import Path
-from typing import Any, Dict, List, Mapping
+from typing import Any, List, Mapping
 
 import pydantic
 import pytest
@@ -80,58 +80,72 @@ def pytest_addoption(parser: pytest.Parser) -> None:
     pow_group.addoption(
         "--ctr-exe",
         metavar="EXE",
+        type=lambda x: x.split(","),
+        default=[],
         help=(
             "Comma-separated list of executables to run the tests with, "
-            "defaults to one/both of 'docker' and 'podman' (depending on which "
-            "is available)"
+            "defaults to using docker, and parameterising with podman if "
+            "available"
         ),
     )
 
 
-def _get_ctr_clients(ctr_exes: List[str], *, all_required: bool) -> List[DockerClient]:
+def _get_ctr_clients(ctr_exes: List[str]) -> List[DockerClient]:
     ctr_clients: List[DockerClient] = []
-    unsupported_ctr_clients: Dict[str, DockerClient] = {}
-    for exe in ctr_exes:
-        client = DockerClient(client_call=[exe])
-        # Check the client is available by running '<client> --version'.
-        try:
-            version_output = subprocess.check_output([exe, "--version"], text=True)
-        except subprocess.CalledProcessError:
-            unsupported_ctr_clients[exe] = client
-            continue
-        # Check the daemon is running with '<client> version'.
-        try:
-            subprocess.check_call([exe, "version"], stdout=subprocess.DEVNULL)
-        except subprocess.CalledProcessError:
-            unsupported_ctr_clients[exe] = client
-            continue
-
-        ctr_clients.append(client)
-
-        # Deduce whether docker or podman from version output.
-        if "docker" in version_output.lower():
-            client.ctr_mgr = "docker"
-        elif "podman" in version_output.lower():
-            client.ctr_mgr = "podman"
+    if not ctr_exes:
+        # Default to using docker, plus podman if available.
+        docker_client = DockerClient()
+        docker_client.ctr_mgr = "docker"
+        ctr_clients.append(docker_client)
+        if (
+            subprocess.run(
+                ["podman", "--version"],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+            ).returncode
+            == 0
+        ):
+            podman_client = DockerClient(client_call=["podman"])
+            podman_client.ctr_mgr = "podman"
+            ctr_clients.append(podman_client)
         else:
-            logger.warning(
-                f"Unrecognised container manager from '{exe} --version', "
-                "assuming docker"
-            )
-            client.ctr_mgr = "docker"
+            logger.warning("Unable to parameterise the tests with podman")
+    else:
+        unsupported_ctr_exes: List[str] = []
+        for exe in ctr_exes:
+            client = DockerClient(client_call=[exe])
+            # Check the client is available by running '<client> --version'.
+            try:
+                version_output = subprocess.check_output([exe, "--version"], text=True)
+            except subprocess.CalledProcessError:
+                unsupported_ctr_exes.append(exe)
+                continue
+            # Check the daemon is running (if required) with '<client> version'.
+            try:
+                subprocess.check_call([exe, "version"], stdout=subprocess.DEVNULL)
+            except subprocess.CalledProcessError:
+                unsupported_ctr_exes.append(exe)
+                continue
 
-    if unsupported_ctr_clients:
-        if all_required:
+            ctr_clients.append(client)
+
+            # Deduce whether docker or podman from version output.
+            if "docker" in version_output.lower():
+                client.ctr_mgr = "docker"
+            elif "podman" in version_output.lower():
+                client.ctr_mgr = "podman"
+            else:
+                logger.warning(
+                    f"Unrecognised container manager from '{exe} --version', "
+                    "assuming docker"
+                )
+                client.ctr_mgr = "docker"
+
+        if unsupported_ctr_exes:
             raise Exception(
-                "'<ctr_exe> --version' failed for the following required exes: "
-                + ",".join(unsupported_ctr_clients)
+                "'<ctr_exe> --version' failed for the following specified exes: "
+                + ",".join(unsupported_ctr_exes)
             )
-        else:
-            logger.warning(
-                "Unable to run with ctr exes: %s", ",".join(unsupported_ctr_clients)
-            )
-        if len(ctr_clients) == 0:
-            raise Exception("No supported ctr exes found")
 
     return ctr_clients
 
@@ -148,16 +162,7 @@ def pytest_configure(config: pytest.Config) -> None:
         "ctr_mgr(MGR, MGR, ...): Container managers supported by the test",
     )
     # Determine supported container managers.
-    if config.getoption("--ctr-exe") is None:
-        config.ctr_clients = _get_ctr_clients(
-            ["docker", "podman"],
-            all_required=False,
-        )
-    else:
-        config.ctr_clients = _get_ctr_clients(
-            config.getoption("--ctr-exe").split(","),
-            all_required=True,
-        )
+    config.ctr_clients = _get_ctr_clients(config.getoption("--ctr-exe"))
 
 
 def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,11 +1,21 @@
+import logging
+import subprocess
 import tempfile
 import time
 from pathlib import Path
+from typing import Any, Dict, List, Mapping
 
 import pydantic
 import pytest
 
-from python_on_whales import docker
+from python_on_whales import DockerClient, docker
+
+logger = logging.getLogger(__name__)
+
+
+#
+# -----------------------------------------------------------------------------
+# Fixtures
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -57,10 +67,130 @@ def swarm_mode():
...
+    config.ctr_clients = _get_ctr_clients(config.getoption("--ctr-exe"))
+
+
+def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
+    if "ctr_client" in metafunc.fixturenames:
+        metafunc.parametrize(
+            "ctr_client",
+            metafunc.config.ctr_clients,
+            ids=[
+                client.client_config.client_call[0]
```
gabrieldemarmiesse, line 173: If there are multiple args, it may be better to do `"-".join(client.client_config.client_call)` no? It's just to be as general as possible.

tests/conftest.py
```diff
@@ -57,10 +67,130 @@ def swarm_mode():
...
+                for client in metafunc.config.ctr_clients
+            ],
         )
+
+
+def pytest_collection_modifyitems(
+    config: pytest.Config, items: List[pytest.Item]
+) -> None:
+    # Find paramaterisations that don't make sense and remove them from
+    # the list of tests to be run.
```
gabrieldemarmiesse, line 183: I'm not sold on this. Pytest already have a mechanism for this and it's the "SKIPPED" status. We should use it. I think it's important that the number of tests collected in the test suite never changes, but that tests can be skipped or deselected using pytest's buit-in features.

tests/python_on_whales/components/test_image.py
```diff
@@ -16,31 +17,32 @@ def test_load_json(json_file):
...
-    docker.image.remove(["busybox:1", "busybox:1.32"])
+def test_image_remove(ctr_client: DockerClient):
+    ctr_client.image.pull("busybox:1", quiet=True)
+    ctr_client.image.pull("busybox:1.32", quiet=True)
+    ctr_client.image.remove(["busybox:1", "busybox:1.32"])
 
 
-def test_image_save_load(tmp_path):
+@pytest.mark.ctr_mgr("docker")
+def test_image_save_load(ctr_client: DockerClient, tmp_path: Path):
```
gabrieldemarmiesse, line 35: Having a custom market may lead to some complexity in case someone wants to understand, debug or change the mechanism. I'd be in favor of something more explicit, like 

```python
@pytest.mark.parametrize("ctr_client", [DOCKER_CLIENT, PODMAN_CLIENT])
def test_image_save_load(ctr_client: DockerClient, tmp_path: Path):
    ...
```
It's easy to debug and understand. DOCKER_CLIENT and PODMAN_CLIENT can be defined somewhere else. It could even be defined with the pytest CLI in case someone doesn't have a docker engine locally, and can use the client with ssh for example (but that's not an immediate need).

gabrieldemarmiesse (changes requested):
Overall, I'm ok with the CLI interface proposed, but I would advocate for using `@pytest.mark.parametrize()`. I'm confident that it's possible from the CLI to skip or deselect some tests. It's even possible that with parametrize, we wouldn't have the implement the CLI interface ourselves and a `-k` would suffice.

You might want to make `DOCKER_CLIENT` and `PODMAN_CLIENT` be a `pytest.param` so that you can choose the id displayed in the test name. See https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids for more info on the subject

LewisGaul:
> Overall, I'm ok with the CLI interface proposed, but I would advocate for using `@pytest.mark.parametrize()`. I'm confident that it's possible from the CLI to skip or deselect some tests. It's even possible that with parametrize, we wouldn't have the implement the CLI interface ourselves and a `-k` would suffice.
> 
> You might want to make `DOCKER_CLIENT` and `PODMAN_CLIENT` be a `pytest.param` so that you can choose the id displayed in the test name. See https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids for more info on the subject

I actually disagree that this is preferable, I very purposefully chose what I have here in favour of using `@pytest.mark.parametrize()`. Let me try to explain why.

My understanding is that this project ideally wants to fully support podman (and possibly nerdctl too?), within the bounds of what we have resources to support (e.g. putting the time in to get the CI running for podman). Therefore, ideally all tests would run with all supported runtimes by default (or at least those that are expected to work).

At a high level, using `@pytest.mark.parametrize()` does not achieve this - it requires *individual testcases* to *opt in* to the parameterisation. It seems to me what we want is for *all testcases that use a container client to be parameterised by default*, with the option of *opting out* (whether being skipped or removed from selection) if unsupported by certain container runtimes.

There are actually two different cases where testcases might need to opt out from a runtime: the runtime does not support the functionality (e.g. docker does not support '`docker pod`' commands or '`docker run --systemd false`'); or POW does not yet support that functionality with the runtime. The reason I went with removing testcases was to cater for the former - in that situation the testcase simply doesn't make sense, so it seemed better for it not to show up as a SKIP which might imply missing functionality in POW. However, you're right that it would make more sense for the latter to be handled by skipping those, rather than removing them (which then serve as a kind of "TODO").

I think we should consider what the desirable behaviour is if a new contributor comes to add new tests. I would expect them to see the pattern in other tests where a `ctr_client` fixture is used, and that we would want their new tests to be *automatically parameterised* with docker and podman, rather than requiring them to *choose to parameterise* using `@pytest.mark.parametrize()`. Then, if the test fails with podman (e.g. in the CI), it forces some thought as to "should this be supported with podman?", and if not then it can be skipped (e.g. using `@pytest.mark.ctr_mgr("docker")`). This has the distinct benefit of being explicit about functionality not yet supported by POW, rather than going unnoticed. As a contributor who favours podman, I don't want to be chasing new testcases to add the parameterisation boilerplate for every testcase!

At a lower implementation-detail level, I'm fully aware `@pytest.mark.parametrize()` would work e.g. with a global `DOCKER_CLIENT` and `PODMAN_CLIENT` as you suggested (although being forced to use global variables is also somewhat undesirable due to wanting to support specifying `--ctr-exe` on the CLI). But it's important to make the right high-level choice before worrying about the implementation - I provided this draft PR to illustrate that my high level proposal *can* be implemented neatly without any use of hacks. In terms of skipping tests instead of removing, this would be a simple change. We could support `@pytest.mark.ctr_mgr("docker", skip=True)` (I've just pushed a demonstration of this) or any other interface - what's your preference?

> Having a custom mark may lead to some complexity in case someone wants to understand, debug or change the mechanism.

Hopefully my code shows it's not actually too complex - this is all supported and documented by pytest, and this feels like a very valid use-case for it. I'm happy to be generous with docstrings/comments in the pytest hooks to explain what's going on, but for most people adding tests they shouldn't need to worry about it, they just see they get a `ctr_client` fixture of type `DockerClient` and shouldn't even need to worry about the fact it causes a parameterisation (but searching for `ctr_client` should lead to my comment that explains what's going on for the curious!).

tests/conftest.py
LewisGaul: `--ctr-exe` currently only supports specifying the exe, not a list of args, but could still change this

tests/conftest.py
LewisGaul: Addressed in my main comment:
> There are actually two different cases where testcases might need to opt out from a runtime: the runtime does not support the functionality (e.g. docker does not support `docker pod` commands or `docker run --systemd false`); or POW does not yet support that functionality with the runtime. The reason I went with removing testcases was to cater for the former - in that situation the testcase simply doesn't make sense, so it seemed better for it not to show up as a SKIP which might imply missing functionality in POW. However, you're right that it would make more sense for the latter to be handled by skipping those, rather than removing them (which then serve as a kind of "TODO").

tests/conftest.py
gabrieldemarmiesse: no worries, i'm just nitpicking here, it's not a big deal

tests/conftest.py
gabrieldemarmiesse: I think we should carify what is expected from a pytest suite:
1) The number of collected items should always be the same, idependently of what is installed on the user's machine or the command line arguments. Here, it changes with the command line arguments, which does not satisfy this condition. This is mainly for visibility purposes and make it easy to reason about the test suite (the list of tests if fixed and the same for all contributors).
2) A test that should not be executed, based on command line arguments (`-k` or custom ones) should be `deselected`. It's what happens when we use `-k` or `-m`. Our custom command line arguments should have a similar effect since they have a similar purpose.
3) A test that can't be executed because a user doesn't have the necessary dependencies on his/her machine should be flagged as SKIPPED. For example, if someone doesn't have podman installed, all podman tests should be marked as SKIPPED, with the corresponding message explaining why.
4) A test that should pass but doesn't due to a bug known in the master branch should be marked as XFAIL, in pytest, it acts as a TODO and allows one person to know what is the expected behavior of the software, even if it doesn't work right now. It's a better documentation than a TODO comment. This can be useful is POW doesn't support something yet but should support it. It can also be used as a form of test-driven development.

If we respect those pytest rules, I don't really mind how it's implemented under the hood.

gabrieldemarmiesse (changes requested):
Some weird stuff is happening, i can't just comment under your messages, I can only make review comments. I hope it's not going to mess up the code review too much

gabrieldemarmiesse:
>My understanding is that this project ideally wants to fully support podman (and possibly nerdctl too?), within the bounds of what we have resources to support (e.g. putting the time in to get the CI running for podman). Therefore, ideally all tests would run with all supported runtimes by default (or at least those that are expected to work).

I disagree with this, but it's just a matter of how it's expressed in the code. I prefer things to be as explicit as possible, even in the default case. I don't have a hard requirement about the parametrize, I just want things to be opt-in. One part of the reason is for explicitness, one other is that it makes migration easier in general to have opt-in options than opt-out (for example if we add nerdctl later, it's possible to progressively add tests that support it instead of having all tests running at once without the opt-out).

>There are actually two different cases where testcases might need to opt out from a runtime: the runtime does not support the functionality (e.g. docker does not support docker pod commands or docker run --systemd false); or POW does not yet support that functionality with the runtime.

If the runtime doesn't support it, if it's the runtime's fault, the test should not exist (it's not even collected) which is what you do here. If POW doesn't support the functionnality, but we'd like POW to support it in the future, the test should be marked as XFAIL and should exist, which is different than what is done in this pull request.

tests/conftest.py
LewisGaul: This all seems reasonable, I'm happy to update to follow these rules.

tests/conftest.py
```diff
@@ -57,10 +67,139 @@ def swarm_mode():
...
+        ),
+    )
+
+
+def _get_ctr_clients(ctr_exes: List[str]) -> List[DockerClient]:
+    ctr_clients: List[DockerClient] = []
+    if not ctr_exes:
+        # Default to using docker, plus podman if available.
+        docker_client = DockerClient()
+        docker_client.ctr_mgr = "docker"
```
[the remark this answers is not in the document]
LewisGaul: > * Let's not add attributes to a class outside an `__init__` function, then the type system of python is lost.

Yeah, is there some way you'd be happy to make this accessible from the `DockerClient` class?

I've previously used a subclass to define this attribute (see [here](https://github.com/LewisGaul/systemd-containers/blob/5993c5f7c55e98f3bf4f2eb0d2ee45fc9dedd4b8/tests/utils.py#L58)), but it felt like that shouldn't be necessary within the project that defines the class itself.

LewisGaul:
> > My understanding is that this project ideally wants to fully support podman (and possibly nerdctl too?), within the bounds of what we have resources to support (e.g. putting the time in to get the CI running for podman). Therefore, ideally all tests would run with all supported runtimes by default (or at least those that are expected to work).
> 
> I disagree with this, but it's just a matter of how it's expressed in the code. I prefer things to be as explicit as possible, even in the default case. I don't have a hard requirement about the parametrize, I just want things to be opt-in. One part of the reason is for explicitness, one other is that it makes migration easier in general to have opt-in options than opt-out (for example if we add nerdctl later, it's possible to progressively add tests that support it instead of having all tests running at once without the opt-out).

I now better understand your perspective, but I still very much think it should be opt-out. Needing to mark tests that don't support one of docker/podman (e.g. with `@pytest.mark.ctr_mgr("docker")`) is a good thing - it makes it clear what functionality is missing under each runtime (there should be more similarity than difference, so more useful to flag the differences).

Given a test that looks like:
```python
def test_image_remove(ctr_client: DockerClient):
    ctr_client.image.pull("busybox:1", quiet=True)
    ctr_client.image.pull("busybox:1.32", quiet=True)
    ctr_client.image.remove(["busybox:1", "busybox:1.32"])
```

and given the fact that the CI runs with podman, I think that test should run with podman. It seems like a gotcha/footgun that it would be as easy as the above (forgetting to use `@pytest.mark.parametrize()`) to miss podman coverage. The `ctr_client` fixture is pytest's dependency-injection approach of defining the client the test should use - there is no reason this should be docker by default, and no reason for the test to care about the fact it might be parameterised.

I would say the more correct way to use pytest is to use fixtures for any resources (where the `DockerClient` instance is the main resource in this case), i.e. this should be done whether used for parameterisation or not. This allows performing checks, controlling initialisation and controlling cleanup - in our case we want it to be parameterised and controlled via a CLI arg.

As listed at https://docs.pytest.org/en/7.3.x/how-to/parametrize.html, there are three ways to parameterise in pytest:
1. `@pytest.mark.parametrize` at the per-testcase level - usually used as a one-off, individual testcase parameterisation
2. At the fixture level (per-module, per-package, ...) - useful for parameterising resources that may be used in multiple testcases
3. `pytest_generate_tests()` hook - needed for more dynamic parameterisation

This case fundamentally does not fall into the first case, this is a resource used by multiple testcases. So, a fixture like this could be used:
```python
@pytest.fixture(scope="session", params=[DOCKER_CLIENT, PYTHON_CLIENT])
def ctr_client(request):
    return request.param
```

However, this gives us no control over:
- Choosing the executables via `--ctr-exe`
- Marking tests as XFAIL or removing those that don't apply

The latter concern could be managed instead with three separate fixtures:
```python
@pytest.fixture(scope="session")
def docker_client():
    return DOCKER_CLIENT

@pytest.fixture(scope="session")
def podman_client():
    return PODMAN_CLIENT

@pytest.fixture(scope="session", params=[DOCKER_CLIENT, PYTHON_CLIENT])
def ctr_client_param(request):
    return request.param
```

such that testcases have to choose whether they want to run with only docker/podman or whether to use the parameterised `ctr_client_param` fixture instead. This would be more explicit about what gets run, but still has the problem of being less explicit about functionality not supported by docker/podman. It also makes supporting `--ctr-exe` more messy/hacky (needing global `DOCKER_CLIENT` and `PODMAN_CLIENT` variables).


> One part of the reason is for explicitness, one other is that it makes migration easier in general to have opt-in options than opt-out (for example if we add nerdctl later, it's possible to progressively add tests that support it instead of having all tests running at once without the opt-out).

I'm glad you mentioned this second point - I think my suggestion fits much better with adding nerdctl. Currently podman and nerdctl are not tested (not run in the CI). I'm doing the work to add podman testing, including getting it running in the CI. At this point of "adding test support", it makes sense to go through and mark any tests that do not work as XFAIL. I'm much less likely to miss any if they all run by default!

In other words, I think the task of adding test support for a new runtime *should* involve explicitly marking any tests that don't work, rather than just getting them running one at a time via explicit parameterisation. In fact, my suggestion of the `--ctr-exe` arg already allows running the tests with nerdctl (as things nerdctl will be treated as docker in terms of what gets removed/XFAILed because it is unrecognised - that's where the `DockerClient.ctr_mgr` attribute comes in).

By the way, I'm very much open to discussion on how the `@pytest.mark.ctr_mgr("docker")` interface should work, e.g. `@pytest.mark.not_supported_by_docker`, or `@pytest.mark.xfail(ctr_mgr="podman")` - there's flexibility here. Similarly the parameterisation of the `ctr_client` fixture could be made more explicit by naming it `ctr_client_param` if preferred.

tests/conftest.py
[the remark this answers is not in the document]
gabrieldemarmiesse: you can just return a list of `Tuple[DockerClient, str]` and work with that.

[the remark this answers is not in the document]
LewisGaul: That's not sufficient, the point is that given a testcase's `ctr_client: DockerClient` I want to know which runtime it corresponds to, to decide whether the testcase is supported or should be marked XFAIL.

gabrieldemarmiesse:
>The ctr_client fixture is pytest's dependency-injection approach of defining the client the test should use - there is no reason this should be docker by default,

I was thinking that all tests should use parametrization, it's just that for the docker-only case, there will be only one parameter. So there wouldn't be a "default". The parametrize doesn't have to be at the function level, it could be at the fixture level and built dynamically.


>However, this gives us no control over:
>Choosing the executables via --ctr-exe
>Marking tests as XFAIL or removing those that don't apply
1) If we parametrize, users can just use `-k` to select or deselect tests. We can also provide markers for the docker tests, so that users can use `-m ` too. Markers are made to do exactly what you are trying to do here.

2) It's possible to apply a xfail in a parametrize: https://docs.pytest.org/en/7.1.x/how-to/skipping.html#skip-xfail-with-parametrize / Removing tests that don't apply just means not adding the corresponding parameter.

I'm afraid we are reinventing the wheel here. Pytest provide ways to dynamically apply parametrize (being at the test level or the function level). Pytest provides a way to mark tests to allow selection and deselection from the CLI, pytest also provide a way to use skip and xfail with parametrize. Let's use the tools that are given to us here. If you don't wish that a user uses parametrize every time a test is written, let's just use the file-level marker `pytestmark` that is given explicitely for this purpose.
It will give you the "by default, all backends are selected" behavior and it's still explicit.

In general, I'd be happier with this PR if we made more use of pytest's features, I believe it would reduce significantly the size of the custom code.

LewisGaul:
> In general, I'd be happier with this PR if we made more use of pytest's features

This doesn't really make sense to me, because my suggested approach makes use of pytest features, I'm not using any hackery. But it seems your preference is to favour code simplicity rather than my strong desire for the tests to have what I'd consider the right default behaviour.

I've been through this process using pytest before, `-m` and `-k`only get you so far, they don't allow the level of control I think we should have here.

Unfortunately I don't think we're going to agree on how this should be approached, and I've lost a lot of motivation to work on this right now. Maybe I'll come back to it another time.

gabrieldemarmiesse:
Don't worry, I can implement the solution when I have the time, I'll see if I can cook something we'll both be satisfied with.

LewisGaul:
One thing I think is essential is that when there's a test that only supports podman, it should not have to be indicated with a parameterisation that includes only podman. That's just inconsistent, it doesn't make sense to parameterise with one thing. Options would be to use a podman fixture, or a mark that indicates the test is only supported by podman, or that indicates lack of support by docker.

PR #524, opened 2024-01-09
:sparkles: Added argument `client_type` to add custom logic later

gabrieldemarmiesse:
@LewisGaul  that should help to do the xfail for the PR https://github.com/gabrieldemarmiesse/python-on-whales/pull/523 

The idea is that it's not required to specify it by a user, but if python-on-whales needs it at some point, we just ask the user to add it to the DockerCLient constructor.

We could call "something version" instead and try to parse the output, but this option is more explicit and faster because it doesn't require a call at every program start.

### DIFF
--- a/python_on_whales/__init__.py
+++ b/python_on_whales/__init__.py
@@ -21,7 +21,7 @@ from .docker_client import DockerClient
 from .exceptions import DockerException
 
 # alias
-docker = DockerClient()
+docker = DockerClient(client_type="docker")
 
 __all__ = [
     "Builder",
--- a/python_on_whales/client_config.py
+++ b/python_on_whales/client_config.py
@@ -5,7 +5,7 @@ import warnings
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Literal, Optional, Union
 
 import pydantic
 
@@ -65,6 +65,7 @@ class ClientConfig:
     compose_project_directory: Optional[ValidPath] = None
     compose_compatibility: Optional[bool] = None
     client_call: List[str] = field(default_factory=lambda: ["docker"])
+    client_type: Literal["docker", "podman", "nerdctl", "unknown"] = "unknown"
     _client_call_with_path: Optional[List[Union[Path, str]]] = None
 
     def get_client_call_with_path(self) -> List[Union[Path, str]]:
--- a/python_on_whales/docker_client.py
+++ b/python_on_whales/docker_client.py
@@ -1,6 +1,6 @@
 import base64
 import warnings
-from typing import List, Optional
+from typing import List, Literal, Optional
 
 from python_on_whales.client_config import ClientConfig, DockerCLICaller
 from python_on_whales.components.buildx.cli_wrapper import BuildxCLI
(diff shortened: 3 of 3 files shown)


LewisGaul:
I'm fine with this, although I think it might be preferable if an approach using `<ctr_exe> --version` could be used.

> We could call "something version" instead and try to parse the output, but this option is more explicit and faster because it doesn't require a call at every program start.

Yeah this is a reasonable concern, although `client_type` could be a property that only runs the command on demand, perhaps caching the result for use by the `version` property too. Also should be very fast to run `<exe> --version`!

python_on_whales/client_config.py
```diff
@@ -65,6 +65,7 @@ class ClientConfig:
     compose_project_directory: Optional[ValidPath] = None
     compose_compatibility: Optional[bool] = None
     client_call: List[str] = field(default_factory=lambda: ["docker"])
+    client_type: Literal["docker", "podman", "nerdctl", "unknown"] = "unknown"
```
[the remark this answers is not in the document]
gabrieldemarmiesse: I know enum is better supported than `Literal` in static type checkers, but it's mostly the fault of type checkers. I expect support will get better in the future.  Using strings is more beginner-friend and works the same with a dynamic type checker like beartype.

gabrieldemarmiesse:
About your comment with "--version", I think we can always add this logic later if we want to. Parsing a output well is hard and not fool proof as the output might change with upgrade. We can always do a call to `--version` if the `client_type` is `unknown` and we want to avoid the user having to type `client_type=...`. For me this PR and a call to `--version` are not incompatible.

python_on_whales/client_config.py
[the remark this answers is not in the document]
LewisGaul: Using enums is more friendly at least to certain users who prefer to use types as guidance for what values are allowed rather than docs/IDE support, but I'm fine with either way.

python_on_whales/docker_client.py
```diff
@@ -102,6 +109,7 @@ def __init__(
         compose_compatibility: Optional[bool] = None,
         client_binary: str = "docker",
         client_call: List[str] = ["docker"],
+        client_type: Literal["docker", "podman", "nerdctl", "unknown"] = "unknown",
```
LewisGaul, line 112: Is this intended as a public attribute of the class? (it would be helpful if it was). If so does it need adding to docs somewhere?

python_on_whales/__init__.py
```diff
@@ -21,7 +21,7 @@
 from .exceptions import DockerException
 
 # alias
-docker = DockerClient()
+docker = DockerClient(client_type="docker")
```
LewisGaul, line 24: If this is because `client_call` defaults to `["docker"]` then why not have this default be set in `DockerClient__init__()`? Your point about the ctr exe possibly being a symlink to something else still stands such that this may be incorrect (unless it actually gets to the point where POW downloads the client).

python_on_whales/docker_client.py
gabrieldemarmiesse: It could be in the future. For the moment, it's not necessary for the goal of this PR.

python_on_whales/client_config.py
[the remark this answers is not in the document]
gabrieldemarmiesse: The codebase uses string everywhere for this use case, there are already litterals here and there. So let's use Literal to be consistent with the rest of the codebase.

python_on_whales/__init__.py
gabrieldemarmiesse: That's true, but then the "unknown" option would never be used, and it has a purpose. It means "as a user I didn't think about it". When people do `from python_on_whales import docker`, we can reasonably assume that they really mean docker, even if you're right, a symlink would be annying here. 

I try to follow the motto "easy things should be easy, hard things should be possible". If there is no "unknown", then it's impossible for the library to tell you that it's assuming incorrectly the type of client.

gabrieldemarmiesse:
Can we merge as-is? We can always improve later on if we want, by making the attribute public or by doing the --version thing.

python_on_whales/__init__.py
LewisGaul: Yeah, that's fine, using `python_on_whales.docker` just comes with the caveat of assuming if you have `docker` on your system it actually is a docker client, I don't have an issue with that.

LewisGaul (approved):

PR #525, opened 2024-01-14
Support podman in test_image.py, improve parameterisation/marking mechanism

LewisGaul:
- Support podman in CI:
  - Add `podman info` output
  - Do not skip tests in CI if docker/podman are not working using new `--no-runtime-skip` CLI arg
  - Deselect podman tests in CI on Windows
- Improve `ctr_client` parameterisation:
  - Automatically apply docker/podman marks (in `pytest_collection_modifyitems()` hook)
  - Remove indirection when parameterising `ctr_client` (no need to import anything from `test_utils`)
  - Simplify code that gets and checks container client (no need for `TestSessionClient` class)
  - Support using `docker_client` and `podman_client` fixtures directly for testcases that do not need parameterising
- Fully parameterise `test_image.py` with podman
  - Some tests updated to support minor differences in podman
  - Other tests marked as xfail for podman for now (note `--runxfail` is useful for bulk fixing xfail tests)
  - Some tests don't make sense to parameterise

More test parameterisation to come (in separate PR)...

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -93,7 +93,7 @@ jobs:
           pip install -r tests/test-requirements.txt
       - name: Run single test
         run: |
-          python -m pytest -vv tests/python_on_whales/components/test_volume.py::test_simple_volume
+          python -m pytest -vv -m "not podman" tests/python_on_whales/components/test_volume.py::test_simple_volume
 
 #  cost too much at the moment.
 #  build-macos:
@@ -118,4 +118,4 @@ jobs:
 #          pip install -r tests/test-requirements.txt
 #      - name: Run all tests
 #        run: |
-#          python -m pytest -vv tests/
+#          python -m pytest -vv -m "not podman" tests/
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -61,7 +61,7 @@ jobs:
           pip install --upgrade ${{ matrix.pydantic-version }}
       - name: Run tests
         run: |
-          python -m pytest -vv --durations=10 tests/python_on_whales/components/${{ matrix.component }}
+          python -m pytest -vv --no-runtime-skip --durations=10 tests/python_on_whales/components/${{ matrix.component }}
 
   build-linux-test-other:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


LewisGaul:
Having solved these concerns I'm now much happier with the setup of explicit parameterisation:
- Can use `docker_client` or `podman_client` directly for tests that don't need parameterising
- Marks for docker/podman get applied directly without needing any indirection in the parameterisation
- Removed the need for the `TestSessionClient` mechanism from `conftest.py`
- Ability to use `pytest.mark.xfail` in parameterisations

gabrieldemarmiesse (changes requested):
While this pull request looks good overall (the way you parametrize the tests are ok, the CLI option looks good too), could you split it into multiple pull requests if that's not too much to ask? I would be able to review much faster and also, it makes it easier to spot bugs/design errors.

For example,, the refactoring you made to change the parametrization can be a standalone pull request. Then you can make 3 other pull requests which depend on that:
* everything related to the registry
* the CLI option
* The rest of the test_image.py file
I'm aware that this requires some more work and it's not fun when writing the code as the code review seems smaller, but this will ensure the project get high quality code review.

You can leave this pull request open as it gives an idea of the final result and this is very useful to have.

I often like to joke with this image, but it's quite true:
super_serious_graph

LewisGaul:
> could you split it into multiple pull requests if that's not too much to ask? I would be able to review much faster and also, it makes it easier to spot bugs/design errors.

While I agree with the sentiment, it's a trade-off between ease of development and ease of reviewing... I've tried to mitigate the size of the review by listing the changes in the description. The real development cost isn't putting it in different PRs, it's having multiple branches based on each other and keeping track of those dependencies, including my work on further parameterisation (I currently have a `podman-tests3` branch based on this `podman-tests2` branch...).

I can go ahead with splitting things up this time as I'm happy to have thorough reviews, but in general this is slowing things down and adding mental overhead for me and that might be something to consider.

Also part of the reason for grouping stuff is it makes it clearer how it all fits together, e.g. you have the whole `test_image.py` file to see how the parameterisation works.

gabrieldemarmiesse:
Well I understand your position, I'm in the same position when I write a lot of code myself. I'm just not confortable reviewing a 400 LOC pull request. The quality of the code review will drop. If you don't want to split it (that's ok), I can split it myself and merge those pull requests.

LewisGaul:
Closing as not intended to be merged, can still be used for reference/comments.

PR #527, opened 2024-01-14
Tidy up `ctr_client` parameterisation/marking mechanism

LewisGaul:
- Improve `ctr_client` parameterisation:
  - Automatically apply docker/podman marks (in `pytest_collection_modifyitems()` hook)
  - Remove indirection when parameterising `ctr_client` (no need to import anything from `test_utils`)
  - Simplify code that gets and checks container client (no need for `TestSessionClient` class)
  - Support using `docker_client` and `podman_client` fixtures directly for testcases that do not need parameterising

First PR factored out of https://github.com/gabrieldemarmiesse/python-on-whales/pull/525

### DIFF
--- a/python_on_whales/test_utils.py
+++ b/python_on_whales/test_utils.py
@@ -4,8 +4,6 @@ from contextlib import contextmanager
 from pathlib import Path
 from typing import List
 
-import pytest
-
 from python_on_whales import client_config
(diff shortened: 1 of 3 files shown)


gabrieldemarmiesse (approved):
Nice job, looks good to me!

PR #529, opened 2024-01-14
Fully parameterise tests with `ctr_client` in `test_image.py`

LewisGaul:
- Fully parameterise `test_image.py` with podman
  - Some tests updated to support minor differences in podman
  - Other tests marked as xfail for podman for now (note `--runxfail` is useful for bulk fixing xfail tests)
  - Some tests don't make sense to parameterise

Third PR factored out of https://github.com/gabrieldemarmiesse/python-on-whales/pull/525


gabrieldemarmiesse (approved):
Thanks! Looks good!

PR #530, opened 2024-01-15
Fully parameterise tests with `ctr_client` in `test_volume.py`

LewisGaul:
Fully parameterise `test_volume.py` with podman.

Some tests marked as xfail for podman for now (note `--runxfail` is useful for bulk fixing xfail tests):
- clone and copy don't work because they rely on buildx, which is a docker thing
- the prune test fails because `--all` isn't supported by podman (my old v20.10.16 docker doesn't support it either)
- the test running `podman volume create --driver local --opt type=tmpfs --opt device=tmpfs --opt o=size=100m,uid=1000` fails with "Error: Volume options size and inodes not supported. Filesystem does not support Project Quota"

Would you like me to create issues to track these, or just use the xfails to track?

### DIFF
--- a/tests/python_on_whales/components/test_volume.py
+++ b/tests/python_on_whales/components/test_volume.py
@@ -4,7 +4,7 @@ from pathlib import Path
 
 import pytest
 
-from python_on_whales import docker
+from python_on_whales import DockerClient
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse (approved):
Looks good!

PR #531, opened 2024-01-15
Fully parameterise tests with `ctr_client` in `test_container.py`

LewisGaul:
Fully parameterise `test_container.py` with podman. Also added a testcase that doesn't rely on mocking for checking podman's systemd mode works, now that this is supported by the test infra.

Some tests marked as xfail for podman for now (note `--runxfail` is useful for bulk fixing xfail tests).

### DIFF
--- a/python_on_whales/components/container/models.py
+++ b/python_on_whales/components/container/models.py
@@ -218,6 +218,7 @@ class ContainerConfig(DockerCamelModel):
     stop_signal: Optional[Union[str, int]] = None
     stop_timeout: Optional[int] = None
     shell: Optional[List[str]] = None
+    systemd_mode: Optional[bool] = None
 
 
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse (approved):
This looks good. I was thinking, what is your opinion on marking all xfails with `strict=True`? With this option, we're garanteed that in future, all tests marked as xfails are really failing. We won't have to periodically run pytest with `--runxfail`. If we do a change that fixes one test, we'll have to remove to xfail for the test to pass.

We can actually use 
```
[pytest]
xfail_strict=true
```
to enable it globally.

gabrieldemarmiesse:
I'll merge this PR, let's continue the discussion about strict xfail here: https://github.com/gabrieldemarmiesse/python-on-whales/issues/353

PR #533, opened 2024-01-15
Parameterise tests with `ctr_client` in `test_network.py`

LewisGaul:
Parameterise `test_network.py` with podman. Note tests using "docker swarm" not parameterised.


LewisGaul:
The issue in the CI is seen on podman v3.4 but not on v4.2 - I'm going to try upgrading the CI's podman version...

PR #534, opened 2024-01-15
Podman upgrade in CI, use xfail_strict and remove xfails that now pass

LewisGaul:
Also checked this unblocks https://github.com/gabrieldemarmiesse/python-on-whales/pull/533. This doesn't actually pin the podman version (it gives us v4.6, which is pretty recent), but I'm not sure how we would do that.


gabrieldemarmiesse (approved):
It's good enough for me and it does fix a lot of xfail. So i'm ok with this, even if there is no pinning.

PR #535, opened 2024-01-15
Move most of the remaining tests to use container client fixtures

LewisGaul:
- The follow commands do not exist with podman (checked with v4.6):
  - `podman config`
  - `podman context`
  - `podman node`
  - `podman plugin`
  - `podman service`
  - `podman stack`
  - `podman swarm`
- The following are not run via parameterisation for now:
  - `podman manifest`
- The following are updated to be parameterised:
  - `podman system`

The above are all updated to use `docker_client` or `ctr_client` such that `--docker-exe` and `--podman-exe` are respected.

### DIFF
--- a/tests/python_on_whales/components/test_config.py
+++ b/tests/python_on_whales/components/test_config.py
@@ -1,41 +1,45 @@
+from pathlib import Path
+
 import pytest
 
-from python_on_whales import docker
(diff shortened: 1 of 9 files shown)


gabrieldemarmiesse (approved):
looks good, thanks!

PR #536, opened 2024-01-15
Tidy up `ReloadableObjectFromJson._fetch_inspect_result_json()`

LewisGaul:
The `ReloadableObjectFromJson._fetch_inspect_result_json()` method implemented for each component is now expected to return an object loaded from JSON, rather than expecting the result from `<ctr_exe> <subcmd> inspect` to always be a list.

This allows removing the hack in `python_on_whales/components/manifest/cli_wrapper.py` (and avoids having to do similar for `podman pod inspect`), which looked like:
```python
    def _fetch_inspect_result_json(self, reference):
        return f'[{run(self.docker_cmd + ["manifest", "inspect", reference])}]'
```

### DIFF
--- a/python_on_whales/client_config.py
+++ b/python_on_whales/client_config.py
@@ -239,14 +239,13 @@ class ReloadableObjectFromJson(ReloadableObject):
         raise NotImplementedError
 
     def _fetch_and_parse_inspect_result(self, reference: str):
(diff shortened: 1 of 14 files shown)


gabrieldemarmiesse (approved):
This is a very good change, indeed the previous implementation was hacky.

PR #538, opened 2024-01-15
Various tidyup for container `cli_wrapper.py`

LewisGaul:
- Correct some type annotations
- Move some formatting helpers to `utils.py` (to be used by the `pod` component too)
- More consistency in constructing commands, e.g. using helpers such as `full_cmd.add_simple_arg()`
- Move `exists()` method to expected place in the classes

### DIFF
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -469,7 +469,7 @@ class ContainerCLI(DockerCLICaller):
         # TODO: fixme
         # full_cmd += ["--pause", str(pause).lower()]
 
-        full_cmd.append(container)
+        full_cmd.append(str(container))
         if tag is not None:
             full_cmd.append(tag)
 
@@ -791,8 +791,8 @@ class ContainerCLI(DockerCLICaller):
 
         full_cmd.add_simple_arg("--workdir", workdir)
 
-        full_cmd.append(str(image))
-        full_cmd.extend(command)
+        full_cmd.append(image)
+        full_cmd += command
         return Container(self.client_config, run(full_cmd), is_immutable_id=True)
 
     def diff(self, container: ValidContainer) -> Dict[str, str]:
@@ -1020,7 +1020,7 @@ class ContainerCLI(DockerCLICaller):
         full_cmd = self.docker_cmd + ["container", "kill"]
 
         full_cmd.add_simple_arg("--signal", format_signal_arg(signal))
-        full_cmd.extend(containers)
+        full_cmd += containers
 
         run(full_cmd)
 
@@ -1792,7 +1792,7 @@ class ContainerCLI(DockerCLICaller):
             containers do not exist.
         """
         containers = to_list(containers)
-        if len(containers) == 0:
+        if containers == []:
             # nothing to do
             return
         full_cmd = self.docker_cmd + ["container", "stop"]
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -39,14 +39,18 @@ from python_on_whales.components.container.models import (
 from python_on_whales.exceptions import NoSuchContainer
 from python_on_whales.utils import (
     ValidPath,
+    ValidPortMapping,
     custom_parse_object_as,
     format_dict_for_cli,
+    format_port_arg,
     format_signal_arg,
     format_time_arg,
+    join_if_not_none,
     removeprefix,
     run,
     stream_stdout_and_stderr,
     to_list,
+    to_seconds,
 )
 
 DockerContainerListFilters = TypedDict(
@@ -292,6 +296,15 @@ class Container(ReloadableObjectFromJson):
             stream,
         )
 
+    def exists(self) -> bool:
+        """Returns `True` if the docker container exists and `False` if it doesn't exists.
+
+        If it doesn't exists, it most likely mean that it was removed.
+
+         See the `docker.container.exists` command for information about the arguments.
+        """
+        return ContainerCLI(self.client_config).exists(self.id)
+
     def export(self, output: ValidPath) -> None:
         """Export this container filesystem.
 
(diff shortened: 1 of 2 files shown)


python_on_whales/components/container/cli_wrapper.py
```diff
@@ -563,31 +560,31 @@ def create(
...
         label_files: List[ValidPath] = [],
         link: List[ValidContainer] = [],
         link_local_ip: List[str] = [],
         log_driver: Optional[str] = None,
         log_options: List[str] = [],
         mac_address: Optional[str] = None,
-        memory: Union[int, str, None] = None,
-        memory_reservation: Union[int, str, None] = None,
-        memory_swap: Union[int, str, None] = None,
+        memory: Optional[Union[int, str]] = None,
```
gabrieldemarmiesse, line 585: I'm not going to block this pull request for this, but note that this is not very useful to change, next python's release, we'll have `int | str | None` instead.

python_on_whales/components/container/cli_wrapper.py
```diff
@@ -1829,14 +1818,15 @@ def unpause(self, x: Union[ValidContainer, List[ValidContainer]], /):
...
         """
-        x = to_list(x)
-        if x == []:
+        containers = to_list(x)
+        if len(containers) == 0:
             # nothing to do
             return
         full_cmd = self.docker_cmd + ["container", "unpause"]
-        full_cmd += x
+        full_cmd.extend(str(c) for c in containers)
```
gabrieldemarmiesse, line 1829: To avoid forgetting the conversion to `str` every time we add something to `full_cmd`, maybe we should make `full_cmd` being a `list[Stringable]` with Stringable being a typing.Protocol that ensures that a method `__str__` is present. Then we can do `str()` when we call the function `run()`. What do you think? That would remove some complexity in the implementation of commands.

gabrieldemarmiesse (changes requested):
very nice cleanup :)

python_on_whales/components/container/cli_wrapper.py
LewisGaul: Yeah sorry this was kind of a style preference change, I personally prefer using `Optional` to indicate arguments where `None` is used to represent "not specified" as opposed to having some other meaning (such as "no timeout"). It's then consistent with the other args that are `Optional[<something>]`, and doesn't look like the `| None` was missed.

I can change these back if you don't like this.

python_on_whales/components/container/cli_wrapper.py
LewisGaul: Yeah, after making some of these changes I realised that `run()` already handles converting all args to strings, so this isn't strictly necessary... One option would be to reflect that in the type signature, although it does have the slight downside of accidentally passing in an object that doesn't have a suitable `__str__()` and it not being obvious when it doesn't work properly. Note that all objects are "stringable" in python.

python_on_whales/components/container/cli_wrapper.py
gabrieldemarmiesse: No worries, I won't block the PR for this, I'm just saying that it's not useful as it's very temporary.

python_on_whales/components/container/cli_wrapper.py
gabrieldemarmiesse: Yeah haha, but it's the same issue here by calling `str()` manually, we can make the mistake and call it on the wrong objects.

python_on_whales/components/container/cli_wrapper.py
LewisGaul: Part of what I was saying is that even with the new syntax available I prefer sticking to `Optional` here, due to how it reads. That is, I'm agreeing with [this perspective](https://discuss.python.org/t/clarification-for-pep-604-is-foo-int-none-to-replace-all-use-of-foo-optional-int/26945/10).

python_on_whales/components/container/cli_wrapper.py
LewisGaul: Well, I guess it's a bit more explicit when done here than in `run()`. But changing `run()` to *not* do it would presumably be difficult at this point. I started making these changes based on it being done in some methods and not others, before I realised `run()` does it automatically. Are you saying you want me to revert these stringify changes?

gabrieldemarmiesse: `run()` already does convert to string, so for me the changes you made are not needed. You can revert back to how it was.

python_on_whales/components/container/cli_wrapper.py
gabrieldemarmiesse: We can't let someone opinion be a standard for an implied meaning, notably because not enough people know about it so it's not going to be applied throughout the codebase, so the meaning in the codebase of `Optional[x]` won't be reliable. Even if the convention comes from a CPython core dev or other. 

As long as it's not in the python documentation or not in a PEP or not in a linter tool to enforce it, we can only assume true what is written in the python docs, that being that `Optional[x] == x | None`. Since the second option is shorter and doesn't require an import, it's going to be preferred in all cases thoughout the project, especially since we can enforce it automatically. We'll likely enforce it with pyupgrade or something similar because memorizing conventions when doing code reviews is a pain.

Long story short, if we adopt the convention proposed here, we have no way of enforce it thoughout the project, unless we force reviewers to force it, and it's very annoying. More automation is always preferable, we'll use pyupgrade for this.

python_on_whales/components/container/cli_wrapper.py
```diff
@@ -795,26 +791,10 @@ def create(
 
         full_cmd.add_simple_arg("--workdir", workdir)
 
-        full_cmd.append(image)
-        full_cmd += command
+        full_cmd.append(str(image))
+        full_cmd.extend(command)
```
gabrieldemarmiesse, line 795:
```suggestion
        full_cmd.append(image)
        full_cmd += command
```

python_on_whales/components/container/cli_wrapper.py
```diff
@@ -1021,13 +1020,13 @@ def kill(
         full_cmd = self.docker_cmd + ["container", "kill"]
 
         full_cmd.add_simple_arg("--signal", format_signal_arg(signal))
-        full_cmd += containers
+        full_cmd.extend(containers)
```
gabrieldemarmiesse, line 1023:
```suggestion
        full_cmd += containers
```
We can't accept diff that changes code style, unless the reason is very strong, otherwise, we'd get PRs to change code style all the time. The first person to write the line chooses the style.

python_on_whales/components/container/cli_wrapper.py
```diff
@@ -1795,21 +1788,16 @@ def stop(
             time: Seconds to wait for stop before killing a container (default 10)
 
         # Raises
-            `python_on_whales.exceptions.NoSuchContainer` if the container does not exists.
+            `python_on_whales.exceptions.NoSuchContainer` if any of the
+            containers do not exist.
         """
         containers = to_list(containers)
-        if containers == []:
+        if len(containers) == 0:
```
gabrieldemarmiesse, line 1795:
```suggestion
        if containers == []:
```
Same here, we don't get enough benefits to justify a small change like this. Otherwise we'd get pull requests with this kind of diff all the time.

gabrieldemarmiesse (changes requested):

gabrieldemarmiesse (approved):
Thanks for the PR, looks good!

Issue #540, opened 2024-01-23
Add `platform` param to `buildx.create()`


rcwbr:
When using the [`buildx.create` function](https://gabrieldemarmiesse.github.io/python-on-whales/sub-commands/buildx/#python_on_whales.components.buildx.cli_wrapper.BuildxCLI.create), it should be possible to specify the platform of the builder container as with the [`docker buildx create --platform` flag](https://docs.docker.com/engine/reference/commandline/buildx_create/#options). This could look something like
```python
buildx.create(
    name="builder-name",
    driver="docker-container",
    platforms=["linux/amd64"],
    use=True,
)
```

gabrieldemarmiesse:
Sure thing. I'm open to such a pull request.

PR #541, opened 2024-01-25
add `platforms` parameter to `buildx.create`

emalihin:
Resolves https://github.com/gabrieldemarmiesse/python-on-whales/issues/540

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "python-on-whales"
-version = "0.68.0"
+version = "0.68.1"
 description = "A Docker client for Python, designed to be fun and intuitive!"
 readme = "README.md"
 license = { file = "LICENSE" }
--- a/python_on_whales/components/buildx/cli_wrapper.py
+++ b/python_on_whales/components/buildx/cli_wrapper.py
@@ -441,6 +441,7 @@ class BuildxCLI(DockerCLICaller):
         context_or_endpoint: Optional[str] = None,
         buildkitd_flags: Optional[str] = None,
         config: Optional[ValidPath] = None,
+        platforms: Optional[List[str]] = None,
         driver: Optional[str] = None,
         driver_options: Dict[str, str] = {},
         name: Optional[str] = None,
@@ -452,6 +453,8 @@ class BuildxCLI(DockerCLICaller):
             context_or_endpoint:
             buildkitd_flags: Flags for buildkitd daemon
             config: BuildKit config file
+            platforms: Comma-separated list of platforms of the form OS/architecture/variant. Ex:
+                `platforms=["linux/amd64", "linux/arm64"]`
             driver: Driver to use (available: [kubernetes docker docker-container])
             driver_options: Options for the driver.
                 e.g `driver_options=dict(network="host")`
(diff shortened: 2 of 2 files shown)


gabrieldemarmiesse:
Thank you for the pull request! It's always nice to see new contributors here :)

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "python-on-whales"
-version = "0.68.0"
+version = "0.68.1"
```
gabrieldemarmiesse, line 7: I manage the versions myself, so you don't need this change.

python_on_whales/components/buildx/cli_wrapper.py
```diff
@@ -441,6 +441,7 @@ def create(
         context_or_endpoint: Optional[str] = None,
         buildkitd_flags: Optional[str] = None,
         config: Optional[ValidPath] = None,
+        platforms: Optional[List[str]] = None,
```
gabrieldemarmiesse, line 444: What would be nice here is to give the users the list of allowed platforms through typing information. We can do this with `typing.Literal`.

If you can, please find the list of available platforms and put them all in the literal:

`  platforms: Optional[List[Literal["linux/amd64", "linux/arm64"]] = None,` (put all allowed literals here)

python_on_whales/components/buildx/cli_wrapper.py
```diff
@@ -465,6 +468,8 @@ def create(
 
         full_cmd.add_simple_arg("--buildkitd-flags", buildkitd_flags)
         full_cmd.add_simple_arg("--config", config)
+        if platforms is not None:
+            full_cmd += ["--platform", ",".join(platforms)]
```
gabrieldemarmiesse, line 472: Can you add a small unit test with multiple platforms to see if it works? Thanks :)

gabrieldemarmiesse (changes requested):

python_on_whales/components/buildx/cli_wrapper.py
emalihin: I borrowed it from [here](https://github.com/gabrieldemarmiesse/python-on-whales/blob/master/python_on_whales/components/buildx/cli_wrapper.py#L224). Is a list of platforms something you would like to maintain?

gabrieldemarmiesse: I don't mind maintaining it, it's very low-effort

emalihin: Do you know where to find a definitive list of supported `os/arch/variant`? Docker docs don't mention it https://docs.docker.com/engine/reference/commandline/buildx_build/#platform

LewisGaul: In general couldn't this change over time, such that old versions of PoW could end up being out of date? I think I'd vote for sticking with strings as with many other fields where we just pass through to the underlying container client.

gabrieldemarmiesse: I believe that it should be better to provide hints to users to get a better UX in the IDE (literals provide great autocompletion in VScode for example). But this can be done in another PR, independently. If it's non-trivial, then let's not block this PR because of it.

gabrieldemarmiesse (approved):

gabrieldemarmiesse:
Many thanks for this PR! It's a cool feature :) And apologies for the delay, I was on holiday.

Issue #543, opened 2024-01-28
`Container.image` could be an `Image` instance rather than string


LewisGaul:
The `Container.image` property is a string representing the image ID, rather than an `Image` instance. This is consistent with all fields of inspect output, which is simply parsed via pydantic, however in the case of the field being a simple ID we could use this to construct a python-on-whales object for the user's convenience.

```diff
     @property
-    def image(self) -> str:
-        return self._get_inspect_result().image
+    def image(self) -> python_on_whales.Image:
+        image_cli = python_on_whales.components.image.cli_wrapper.ImageCLI(self.client_config)
+        image_id = self._get_inspect_result().image
+        return image_cli.inspect(image_id)
```

The problem is backwards compatibility... Perhaps something for v1.0?

The same consideration exists for the new `Container.pod` and `Pod.containers` properties, see https://github.com/LewisGaul/python-on-whales/pull/4/commits/6ffcf45813006e90377f3054495ec16dfd879276.

gabrieldemarmiesse:
Doing a `docker.image.inspect()` has a runtime cost. If we do it, we'll need to use the lazy loading features of python-on-whales. 

I'll need to think about it, there are qui a few area that this can affect, because we could argue that the same could be done for volumes, networks, etc

PR #544, opened 2024-01-29
Implement support for podman pods

LewisGaul:
This feature only applies for podman, not docker.
- Support `podman create --pod <pod>` and `podman run --pod <pod>` (`ContainerCLI.create(pod=...)` and `ContainerCLI.run(pod=...)`)
- Support `podman inspect <ctr>` including the pod if applicable (`Container.pod` property)
- Support pod subcommand/component, `podman pod ...` (`Pod`, `PodCLI`, and corresponding inspect data model objects)
- Add tests using podman client
- Update docs (checked locally)

Questions/notes:
- How should we encorporate this in the docs, since the commands are current listed as "docker \<subcmd>" but docker does not support "docker pod"?
- Following response to the above, I have added "podman pod" subcmd on the docs, and changed "Docker objects" to just "Objects" to allow inclusion of "Pods".
- I have implemented `Container.pod` to return the string pod ID rather than the `Pod` object - see https://github.com/gabrieldemarmiesse/python-on-whales/issues/543 for discussion.
- Is it the right choice to support args like `pod_id_file` in `PodCLI.create()`?
- Deferring implementation of `podman pod stats` and `podman pod top`.
- Deferring implementation of support for `--latest` and `--all` (for `podman pod {start,restart,stop,kill,pause,unpause}` and similar for podman containers) following comments in this PR, tracked by https://github.com/gabrieldemarmiesse/python-on-whales/issues/588.

TODO:
- [x] Docs update
- [x] More tests needed
- [x] Docstring for `PodCLI.create()`

Follow-up:
- Add the new `test_pod.py` to required PR checks (requires repo admin)
- Skip tests that aren't supported on older podman versions, depends on https://github.com/gabrieldemarmiesse/python-on-whales/issues/591


LewisGaul:
@gabrieldemarmiesse any thoughts on what needs doing here, with regards to docs and tests?

gabrieldemarmiesse:
>How should we encorporate this in the docs, since the commands are current listed as "docker <subcmd>" but docker does not support "docker pod"?

We can choose the name of web pages, do instead of naming it "docker ..." like the other commands, we could just name it "podman pod"

> I have implemented Container.pod to return the string pod ID rather than the Pod object 

Good call, I still haven't decided what to do here.

> Is it the right choice to support --latest and --all to podman pod {start,restart,stop,kill,pause,unpause}?

I think so yes. We want python-on-whales to be familiar to users who started with the CLI (in this case the podman CLI).

The rest of the review will be done, no worries, I'm just finding the time here and there

python_on_whales/components/pod/cli_wrapper.py
```diff
@@ -0,0 +1,778 @@
...
+        """
+        return PodCLI(self.client_config).kill(self, signal=signal)
+
+    def logs(
+        self,
+        *,
+        container: Optional[
+            python_on_whales.components.container.cli_wrapper.ValidContainer
+        ] = None,
+        names: bool = False,
```
gabrieldemarmiesse, line 176:
```suggestion
        container: Optional[
            python_on_whales.components.container.cli_wrapper.ValidContainer
        ] = None,
        *,
        names: bool = False,
```

I think we should allow `my_pod.logs(my_container_name)`, it doesn't necessitate additional information to be understood.

python_on_whales/components/pod/cli_wrapper.py
```diff
@@ -0,0 +1,778 @@
...
+        """
+        return PodCLI(self.client_config).stop(self, time=time)
+
+
+ValidPod = Union[str, Pod]
+
+
+class PodCLI(DockerCLICaller):
+    def __init__(self, client_config: ClientConfig):
+        super().__init__(client_config)
```
gabrieldemarmiesse, line 255: Is this needed? That's the default behavior of `DockerCLICaller` no?

python_on_whales/components/pod/cli_wrapper.py
```diff
@@ -0,0 +1,778 @@
...
+        labels: Dict[str, str] = {},
+        label_files: List[ValidPath] = [],
+        mac_address: Optional[str] = None,
+        memory: Union[int, str, None] = None,
+        networks: List[
+            python_on_whales.components.network.cli_wrapper.ValidNetwork
+        ] = [],
+        network_aliases: List[str] = [],
+        no_hosts: bool = False,
+        pid: Optional[str] = None,
```
gabrieldemarmiesse, line 291: Shouldn't this be an int?

python_on_whales/components/pod/cli_wrapper.py
```diff
@@ -0,0 +1,778 @@
...
+        security_options: List[str] = [],
+        share: List[str] = [],
+        shm_size: Optional[Union[int, str]] = None,
+        subgidname: Optional[str] = None,
+        subuidname: Optional[str] = None,
+        sysctl: Dict[str, str] = {},
+        uidmaps: List[Tuple[int, int, int]] = [],
+        userns: Optional[str] = None,
+        uts: Optional[str] = None,
+        volumes: Optional[List[VolumeDefinition]] = [],
```
gabrieldemarmiesse, line 304:
```suggestion
        volumes: List[VolumeDefinition] = [],
```
The `None` is not checked in the code

python_on_whales/components/pod/cli_wrapper.py
```diff
@@ -0,0 +1,778 @@
...
+        ]
+
+    def logs(
+        self,
+        pod: ValidPod,
+        *,
+        container: Optional[
+            python_on_whales.components.container.cli_wrapper.ValidContainer
+        ] = None,
+        names: bool = False,
```
gabrieldemarmiesse, line 464:
```suggestion
        /,
        container: Optional[
            python_on_whales.components.container.cli_wrapper.ValidContainer
        ] = None,
        *,
        names: bool = False,
```
To be consistent with the other comment.

python_on_whales/components/pod/cli_wrapper.py
```diff
@@ -0,0 +1,778 @@
...
+        iterator = stream_stdout_and_stderr(full_cmd)
+
+        if stream:
+            return iterator
+        else:
+            return "".join(x[1].decode() for x in iterator)
+
+    def pause(
+        self,
+        x: Union[ValidPod, Sequence[ValidPod]] = (),
```
gabrieldemarmiesse, line 535:
```suggestion
        x: Union[ValidPod, Sequence[ValidPod]],
```
What's the use case for a default value?

python_on_whales/components/pod/cli_wrapper.py
```diff
@@ -0,0 +1,778 @@
...
+        run(full_cmd)
+
+    ps = list
+
+    def remove(
+        self,
+        x: Union[ValidPod, Sequence[ValidPod]] = (),
+        /,
+        *,
+        all: bool = False,
```
gabrieldemarmiesse, line 574: I wonder if there is an elegant way to say that `all` can't be used at the same time as `x`. Maybe with overloads? Or maybe with a special sentinel value? I'm always in favor of having the IDE help as much as possible. We also might want to verify that both all and x are not set at the same time and raise an error if this is the case.

python_on_whales/components/pod/cli_wrapper.py
```diff
@@ -0,0 +1,778 @@
...
+
+        run(full_cmd)
+
+    def stats(
+        self,
+        x: Union[ValidPod, Sequence[ValidPod]] = (),
+        /,
+        *,
+        all: bool = False,
+        latest: bool = False,
```
gabrieldemarmiesse, line 668: Same issue here. I wonder if there is a way to say that only one argument must be used. But it's less of an issue since it's not implemented yet.

python_on_whales/components/pod/cli_wrapper.py
```diff
@@ -0,0 +1,778 @@
...
+        stats_output = run(full_cmd)
+        return [PodStats(json.loads(s)) for s in stats_output.splitlines()]
+
+    def stop(
+        self,
+        x: Union[ValidPod, Sequence[ValidPod]] = (),
+        /,
+        *,
+        all: bool = False,
+        latest: bool = False,
```
gabrieldemarmiesse, line 707: Here too, only one argument must be used, I wonder how we can add this information in the signature.

tests/python_on_whales/components/test_pod.py
gabrieldemarmiesse, line 1: The tests are fine, Would it be possible to add a few more? The things that seems to be missing are
* More tests with arguments in the `create` method. Especially the ones that require special parsing.
* Tests with `all=True` and `latest=True`.

gabrieldemarmiesse (changes requested):
Many thanks for the pull request!

gabrieldemarmiesse:
> Is it the right choice to support args like pod_id_file in PodCLI.create()?
I don't see what is special about it. Maybe I'm missing something?

gabrieldemarmiesse:
Food for thought, I wonder if we should have a sentinel value for "all objects" that we can use in this PR but also as default value in some other places. It might seems strange to have `containers=None` mean all containers (since it's the default value). We could also have a sentinel for `latest`.

python_on_whales/components/pod/cli_wrapper.py
LewisGaul: `podman.pod.pause(all=True)` or `podman.pod.pause(latest=True)`

python_on_whales/components/pod/cli_wrapper.py
```diff
@@ -0,0 +1,778 @@
...
+        full_cmd.extend([str(p) for p in pods])
+
+        run(full_cmd)
+
+    def prune(self) -> None:
+        """Remove pods that are not running."""
+        full_cmd = self.docker_cmd + ["pod", "prune", "--force"]
+        run(full_cmd)
+
+    ps = list
```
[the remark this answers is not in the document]
LewisGaul: If you prefer. We don't group things like `pause()` and `unpause()`

python_on_whales/components/pod/cli_wrapper.py
LewisGaul: I don't think we should make the `pod` argument positional-only, there's nothing wrong with a user doing `podman.pod.logs(container="my-ctr", pod="my-pod")`, unlike other methods which accept one or more pods so neither "pod" nor "pods" are always correct.

python_on_whales/components/pod/cli_wrapper.py
LewisGaul: There's an inconsistency in the CLI here... The following are used for controlling namespaces: `--cgroupns` (cgroup), `--network` (net), `--pid` (pid), `--userns` (user).
```
$podman pod create --help | grep -- --pid
      --pid string                    PID namespace to use
```
This will be clearer when I've added the method docstring :)

python_on_whales/components/pod/cli_wrapper.py
LewisGaul: I'll have a go at using overloads - might be a bit verbose but I can see it should be helpful. A sentinel value might be nice (could just have a global `ALL = object()` and `LATEST = object()`) but perhaps a bit less obvious than just mirroring the CLI. Probably best to have a separate issue for that if you want to track the idea.

LewisGaul:
> > Is it the right choice to support args like pod_id_file in PodCLI.create()?
> 
> I don't see what is special about it. Maybe I'm missing something?

Only that it's probably easier to get the ID of the pod when using PoW than when calling the CLI so less use for having podman write it to a file for you (tbh this seems like a slightly odd CLI arg to me). But happy to just leave it for consistency with the CLI.

python_on_whales/components/pod/cli_wrapper.py
LewisGaul: I've removed support for `all` and `latest` in this PR since that's what a lot of the comments are about, so this can be discussed and addressed separately (https://github.com/gabrieldemarmiesse/python-on-whales/issues/588). E.g. there's also equivalent flags on some `podman container` commands.

python_on_whales/components/pod/cli_wrapper.py
LewisGaul: deferred

python_on_whales/components/pod/cli_wrapper.py
LewisGaul: deferred

python_on_whales/components/pod/cli_wrapper.py
LewisGaul: resolved by removing `all` and `latest`

tests/python_on_whales/components/test_pod.py
LewisGaul: I've added tests for some of the different `pod.create()` arguments. Some of them are a bit tricky since they only work on certain hosts, e.g. cgroups v2, others I'm just not exactly sure how to test.

I removed `all` and `latest` for now...

LewisGaul:
@gabrieldemarmiesse This should be ready for review now

gabrieldemarmiesse (approved):
That's great work, many thanks! Since there are so many lines of code, it's quite possible that I missed a few things but I believe we can iterate a bit on the code before creating a new release. Let's welcome pods into Python-on-whales!

PR #549, opened 2024-02-09
Fix `client.container.start(attach=True)`

LewisGaul:
Currently output is captured by PoW when starting a container attached, which is inconsistent behaviour compared to `client.container.attach()`.

### DIFF
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -1772,9 +1772,10 @@ class ContainerCLI(DockerCLICaller):
         if stream:
             return stream_stdout_and_stderr(full_cmd)
         elif attach:
-            return run(full_cmd)
+            run(full_cmd, tty=True)
+            return None
         else:
-            run(full_cmd)
+            return run(full_cmd)
 
     def stats(
         self,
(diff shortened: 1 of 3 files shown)


python_on_whales/components/container/cli_wrapper.py
```diff
@@ -1726,9 +1726,10 @@ def start(
         if stream:
             return stream_stdout_and_stderr(full_cmd)
         elif attach:
-            return run(full_cmd)
+            run(full_cmd, tty=True)
+            return None
```
gabrieldemarmiesse, line 1730: To me, this is indeed a breaking change, but it makes sense. Furthermore, if a user is interested in stdout, he can either use stream or the normal call. For me this is acceptable (we're not in the v1.0 yet). Feel free to  adapt the unit tests. I'll add a big fat warning in the release notes. Also, can you add a warning in the docstring saying that the bahavior changed? Thanks!

gabrieldemarmiesse (changes requested):

python_on_whales/components/container/cli_wrapper.py
LewisGaul: Sounds good, thanks!

LewisGaul:
Closing because I think this needs a bit more high-level thought... Tracked by https://github.com/gabrieldemarmiesse/python-on-whales/issues/550.

PR #554, opened 2024-02-15
Add `detach_keys` argument for `start` and `exec`

jhc4318:
- [`docker start`](https://docs.docker.com/engine/reference/commandline/container_start/#options)
- [`docker exec`](https://docs.docker.com/engine/reference/commandline/container_exec/#options)
Also supported with `podman`.

UT just checks the args are passed correctly.

### DIFF
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -389,7 +389,11 @@ class Container(ReloadableObjectFromJson):
         )
 
     def start(
-        self, attach: bool = False, detach_keys: Optional[str] = None, interactive: bool = False, stream: bool = False
+        self,
+        attach: bool = False,
+        detach_keys: Optional[str] = None,
+        interactive: bool = False,
+        stream: bool = False
     ) -> Union[None, str, Iterable[Tuple[str, bytes]]]:
         """Starts this container.
 
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -846,7 +846,7 @@ class ContainerCLI(DockerCLICaller):
             command: The command to execute.
             detach: if `True`, returns immediately with `None`. If `False`,
                 returns the command stdout as string.
-            detach_keys: Key sequence to detach from the container.
+            detach_keys: Override the key sequence for detaching a container.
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse:
Hi, there seems to be conflicts in this pull request, can you fix them? Thank you very much :)

python_on_whales/components/container/cli_wrapper.py
```diff
@@ -387,14 +389,20 @@ def remove(self, force: bool = False, volumes: bool = False) -> None:
         )
 
     def start(
-        self, attach: bool = False, interactive: bool = False, stream: bool = False
+        self,
+        attach: bool = False,
+        detach_keys: Optional[str] = None,
+        interactive: bool = False,
+        stream: bool = False,
```
gabrieldemarmiesse, line 396:
```suggestion
        interactive: bool = False,
        stream: bool = False,
        detach_keys: Optional[str] = None,
```
To avoid breaking backward compatibility. We'll disally positional arguments soon anyway for this method, and the order won't matter then. But for now, we don't want any silent breakage.

python_on_whales/components/container/cli_wrapper.py
```diff
@@ -827,6 +835,7 @@ def execute(
         container: ValidContainer,
         command: List[str],
         detach: bool = False,
+        detach_keys: Optional[str] = None,
```
gabrieldemarmiesse, line 838: Same here.

python_on_whales/components/container/cli_wrapper.py
```diff
@@ -1699,6 +1710,7 @@ def start(
         self,
         containers: Union[ValidContainer, List[ValidContainer]],
         attach: bool = False,
+        detach_keys: Optional[str] = None,
```
gabrieldemarmiesse, line 1713: Same here

gabrieldemarmiesse (changes requested):

python_on_whales/components/container/cli_wrapper.py
jhc4318: Makes sense, I'll make the relevant changes!

gabrieldemarmiesse (approved):
Many thanks! Looks good!

Issue #557, opened 2024-02-27
`DockerClient.image.exists` using `podman` does not work


jhc4318:
`image.exists` just calls `image.inspect` under the covers and parses the error output for `no such image` to determine if the image exists or not. https://github.com/gabrieldemarmiesse/python-on-whales/blob/4433a0227b90974f9e0c7090383deacedf8d01c6/python_on_whales/utils.py#L156

This works with `docker`
```
>docker image inspect hello
[]
Error response from daemon: No such image: hello:latest
```

PR #558, opened 2024-02-27
Fix`podman.image.exists`

jhc4318:
As described in this issue https://github.com/gabrieldemarmiesse/python-on-whales/issues/557

I don't see any tests for `utils.py` which cover checking stderr for the different messages and raising different exceptions, however happy to add if you feel it's necessary or would be useful to begin building up some tests now.


gabrieldemarmiesse (approved):
Many thanks for the fix and the cleanup. I wish we could use exit codes for this, but it seems CLIs don't bother with multiple exit code...

PR #565, opened 2024-03-11
fix repr return for Tasks

### LINKED ISSUE #555: Bug ?
Sorry if I'm generating noise, but I believe this line was supposed to say "Task" and not "Image" ; https://github.com/gabrieldemarmiesse/python-on-whales/blob/4ed016dcc8fa566a5bb1429c5cc5c73cd2864bc6/python_on_whales/components/task/cli_wrapper.py#L93

anesmemisevic:
### Fixes #555, 
- repr now returns the valid type


gabrieldemarmiesse (approved):
Many thanks!

Issue #570, opened 2024-03-21
Function  `docker.compose.down`  does not accept optional list of service names


MisterOwlPT:
Hi 😃 

The Docker _Compose_ CLI allows the command `down` to be passed a list of service names  (similar to other commands `up` and `rm`).

```bash
$ docker compose down --help

Usage:  docker compose down [OPTIONS] **[SERVICES]**
Stop and remove containers, networks
...
```

I noticed that `python_on_whales` does not currently support this feature, which I personally need for a personal use case.
Calling `docker.compose.down()` simply stops _all_ declared service containers. 

If you agree, I can submit a PR with a logic similar to [this](https://github.com/gabrieldemarmiesse/python-on-whales/blob/master/python_on_whales/components/compose/cli_wrapper.py#L1007) such that if service names are provided, then only the related containers are stopped. Otherwise, if no list of service names is provided, _all_ service containers are stopped.

gabrieldemarmiesse:
Thank you for opening this feature request! When I made the first iteration of python-on-whales, this was not available. Feel free to make a pull request, I'll take a look at it.

gabrieldemarmiesse:
Closing as fixed by #571

PR #571, opened 2024-03-21
Function docker.compose.down does not accept optional list of service names #570 

MisterOwlPT:
Hi 😃 

Here follow the proposed changes regarding issue #570.
Now the following calls are possible:

```python
from python_on_whales import docker

# ...

docker.compose.down(services=[])  # no-op - does nothing
docker.compose.down(services="a")  # stops service "a"
docker.compose.down(services=["b", "c"])  # stops services "b" and "c"
docker.compose.down()  # default - stops all services
```

- Unit tests included
- Documentation updated

### DIFF
--- a/tests/python_on_whales/components/test_compose.py
+++ b/tests/python_on_whales/components/test_compose.py
@@ -651,7 +651,7 @@ def test_config_complexe_compose():
     docker = DockerClient(compose_files=[compose_file], compose_compatibility=True)
     config = docker.compose.config()
 
-    assert config.services["my_service"].build.context == Path(
+    assert config.services["my_service"].build.context == (
         PROJECT_ROOT / "tests/python_on_whales/components/my_service_build"
     )
     assert config.services["my_service"].image == "some_random_image"
--- a/tests/python_on_whales/components/test_compose.py
+++ b/tests/python_on_whales/components/test_compose.py
@@ -5,6 +5,7 @@ import time
 from datetime import datetime, timedelta
 from os import makedirs, remove
 from pathlib import Path
+from unittest import Mock, patch
 
 import pytest
 import pytz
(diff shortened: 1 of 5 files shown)


MisterOwlPT:
Hi @gabrieldemarmiesse,

I encountered a problem  while submitting this PR. In essence, the tests I implemented are passing on my local machine but are failing inside the runner.

Initially I thought the tests were failing due to some unexpected service and container naming inside the runner and committed some extra "safety" changes that ended up not solving anything (I can revert these later). Upon investigating a bit further, I found that the CI/CD pipeline configuration [enforces a specific version of the Docker Compose plugin](https://github.com/gabrieldemarmiesse/python-on-whales/blob/master/scripts/download-docker-plugins.sh#L6) (`v2.15.1`), which does not support service names on the `down` command.

For sanity check I tried running my tests on a brand new virtual machine provisioned with the exact same version of the Compose plugin as the runner.

```bash
# Inside VM - using same plugin version as runner
$ docker compose version
Docker Compose version v2.15.1

# Plugin does not support SERVICES
$ docker compose down --help
Usage:  docker compose down [OPTIONS]
Stop and remove containers, networks
```

Running my tests on said VM I was able to replicate the [error](https://github.com/gabrieldemarmiesse/python-on-whales/actions/runs/8376629016/job/22936751979#step:5:224) (check below).

```bash
# Extracted from pytest command
Container alpine  Running
Container busybox  Running
unknown command "busybox" for "docker compose down"
```

Here is the output of `docker info` on my machine. You can see I am using a newer version of said plugin:

```bash
$ docker info
Client: Docker Engine - Community
 Version:    25.0.4
 ...
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.13.0
  compose: Docker Compose (Docker Inc.)
    Version:  v2.24.7
```

As this dependency constraint is beyond my control, I wanted to bring it to your attention.
Is there a particular reason for using that version? Can it be upgraded?

gabrieldemarmiesse:
You can upgrade the version in the CI to the latest one. We are pinning the version to make sure that we don't end up with failing tests because of automatic upgrade. So our CI is reproducible. Feel free to change the version in the bash file.

MisterOwlPT:
Thanks for the feedback!
I updated to the _latest_ version (`v2.26.0`).

I also add to fix test `test_config_complexe_compose` as it was affected by the plugin version update (command `config` prints distinct outputs for field `build.context`).

```bash
# Using old version v2.15.1
$ docker compose -f tests/python_on_whales/components/complexe-compose.yml config
name: components
services:
  my_service:
    build:
      context: my_service_build
      dockerfile: Dockerfile
```

```bash
# Using latest version v2.26.0
$ docker compose -f tests/python_on_whales/components/complexe-compose.yml config
name: components
services:
  my_service:
    build:
      context: /home/ubuntu/python-on-whales/tests/python_on_whales/components/my_service_build
      dockerfile: Dockerfile
```

Now the tests are passing 😃

tests/python_on_whales/components/test_compose.py
```diff
@@ -563,7 +651,9 @@ def test_config_complexe_compose():
     docker = DockerClient(compose_files=[compose_file], compose_compatibility=True)
     config = docker.compose.config()
 
-    assert config.services["my_service"].build.context == Path("my_service_build")
+    assert config.services["my_service"].build.context == Path(
+        PROJECT_ROOT / "tests/python_on_whales/components/my_service_build"
+    )
```
gabrieldemarmiesse, line 656:
```suggestion
    assert config.services["my_service"].build.context == PROJECT_ROOT / "tests/python_on_whales/components/my_service_build"
```
It's already a Path, there is no need to convert it to a path again.

gabrieldemarmiesse (changes requested):
Very nice pull request, a pleasure to read. Just a small line to simplify and we're merging :)

tests/python_on_whales/components/test_compose.py
MisterOwlPT: Whoops 😅  
Fixed. I left a multi-line declaration as I find it more readable.

Thanks a lot for your feedback!

gabrieldemarmiesse (approved):

PR #572, opened 2024-03-22
Improve typing for `DockerClient.compose.config`

einarwar:
Added overloads to make typing more exact by evaluating the truthiness of `return_json`


gabrieldemarmiesse (approved):
Many thanks! This should help IDEs :)

PR #574, opened 2024-03-26
Fix time argument formatting to include time zone

kamalmarhubi:
In local testing, I saw that `Container.logs` misbehaved when passing the `since` parameter. I believe this is because my time zone is UTC-4, and so asking for something like

    container.logs(since=container.state.started_at)

would return no logs as it was effectively asking for logs from the future:
- container.state.started_at parses the inspect output into a zone-aware datetime; the timezone in inspect's output is always UTC (`Z`)
- the time formatting in `format_time_for_docker` strips the time zone rather than converts to local time
- the docker CLI treats its `--since` argument as being in local time

I instrumented the Popen call to see what arguments were passed to the docker CLI. Taking the exact same command and appending a `Z` to the time to force the correct time zone printed logs as expected.

### DIFF
--- a/python_on_whales/utils.py
+++ b/python_on_whales/utils.py
@@ -323,7 +323,7 @@ def format_time_arg(time_object):
 
 def format_time_for_docker(time_object: Union[datetime, timedelta]) -> str:
     if isinstance(time_object, datetime):
-        return time_object.strftime("%Y-%m-%dT%H:%M:%S")
+        return time_object.isoformat()
     elif isinstance(time_object, timedelta):
         return f"{time_object.total_seconds()}s"
 


gabrieldemarmiesse:
Many thanks for the thorough bug hunting! That's awesome! From what you described, it sounds like a good fix. I'll take a look in details and give you a review in the next few days :) I also wonder if this kind of bug can occur somewhere else in the codebase.

gabrieldemarmiesse (approved):
Thanks! Let's merge it in master. Can you try it out with `pip install git+https://github.com/gabrieldemarmiesse/python-on-whales.git` and tell me if it works?

kamalmarhubi:
@gabrieldemarmiesse yup I removed my monkey patch to test with current master and it works, thanks!

PR #576, opened 2024-04-08
Support podman's `--preserve-fds` arg to container run/exec

LewisGaul:
Tested manually, but should probably add a testcase too... need to figure out the best way to do so.

- [x] Add testcase

### DIFF
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -844,6 +844,7 @@ class ContainerCLI(DockerCLICaller):
         workdir: Optional[ValidPath] = None,
         stream: bool = False,
         detach_keys: Optional[str] = None,
+        preserve_fds: Optional[int] = None,
     ) -> Union[None, str, Iterable[Tuple[str, bytes]]]:
         """Execute a command inside a container
 
@@ -860,6 +861,9 @@ class ContainerCLI(DockerCLICaller):
                 to allow communication with the parent process.
                 Currently only works with `tty=True` for interactive use
                 on the terminal.
+            preserve_fds: The number of additional file descriptors to pass
+                through to the container. Only supported by podman.
+
             privileged: Give extended privileges to the container.
             tty: Allocate a pseudo-TTY. Allow the process to access your terminal
                 to write on it.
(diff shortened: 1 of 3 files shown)


python_on_whales/components/container/cli_wrapper.py
```diff
@@ -860,6 +861,9 @@ def execute(
                 to allow communication with the parent process.
                 Currently only works with `tty=True` for interactive use
                 on the terminal.
+            preserve_fds: The number of additional file descriptors to pass
+                through to the container. Only supported by podman.
+
```
gabrieldemarmiesse, line 866:
```suggestion
                through to the container. Only supported by podman.
```

python_on_whales/components/container/cli_wrapper.py
```diff
@@ -1725,12 +1739,17 @@ def run(
                     "It's not possible to stream and detach a container at "
                     "the same time."
                 )
+
+        if preserve_fds:
+            pass_fds = range(3, 3 + preserve_fds)
```
gabrieldemarmiesse, line 1744: I wouldn't mind an explanation in a code comment for this line. That should speed up the comprehension for future readers

tests/python_on_whales/components/test_container.py
```diff
@@ -218,6 +218,17 @@ def test_create_with_systemd_mode(podman_client: DockerClient):
         assert container.config.systemd_mode is True
 
 
+def test_run_with_preserve_fds(podman_client: DockerClient):
+    read_fd, write_fd = os.pipe()
+    with podman_client.container.run(
+        "ubuntu",
+        ["bash", "-c", f"echo foobar >&{write_fd}"],
+        detach=True,
+        preserve_fds=write_fd - 2,
```
gabrieldemarmiesse, line 227: Same here

gabrieldemarmiesse (changes requested):
Thanks for the pull request. I'm not super familiar with file descriptors. I guess we can assume that a reader knows they're represented by integers. We should just add a few comments here and there to explain the magic numbers and we should be good to go

python_on_whales/components/container/cli_wrapper.py
gabrieldemarmiesse: There is an empty extra line, this will break the docs formatting

LewisGaul: oops

gabrieldemarmiesse (approved):
Looks good, a high quality PR once again :)

PR #579, opened 2024-04-17
Added timezone argument to run and create

tjd78:
- Added `tz` to the arguments for `run()` and `create()` as per the arguments supported in docker and podman:
https://docs.podman.io/en/stable/markdown/podman-create.1.html#tz-timezone
- Added testcases to demonstrate the change working as expected.

### DIFF
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -626,6 +626,7 @@ class ContainerCLI(DockerCLICaller):
         systemd: Optional[Union[bool, Literal["always"]]] = None,
         tmpfs: List[ValidPath] = [],
         tty: bool = False,
+        tz: Optional[str] = None,
         ulimit: List[str] = [],
         user: Optional[str] = None,
         userns: Optional[str] = None,
@@ -791,6 +792,7 @@ class ContainerCLI(DockerCLICaller):
         full_cmd.add_simple_arg("--systemd", systemd)
         full_cmd.add_args_list("--tmpfs", tmpfs)
         full_cmd.add_flag("--tty", tty)
+        full_cmd.add_simple_arg("--tz", tz)
         full_cmd.add_args_list("--ulimit", ulimit)
 
         full_cmd.add_simple_arg("--user", user)
@@ -1387,6 +1389,7 @@ class ContainerCLI(DockerCLICaller):
         systemd: Optional[Union[bool, Literal["always"]]] = None,
         tmpfs: List[ValidPath] = [],
         tty: bool = False,
+        tz: Optional[str] = None,
         ulimit: List[str] = [],
         user: Optional[str] = None,
         userns: Optional[str] = None,
@@ -1718,6 +1721,7 @@ class ContainerCLI(DockerCLICaller):
         full_cmd.add_simple_arg("--systemd", systemd)
         full_cmd.add_args_list("--tmpfs", tmpfs)
         full_cmd.add_flag("--tty", tty)
+        full_cmd.add_simple_arg("--tz", tz)
         full_cmd.add_args_list("--ulimit", ulimit)
 
         full_cmd.add_simple_arg("--user", user)
--- a/tests/python_on_whales/components/test_container.py
+++ b/tests/python_on_whales/components/test_container.py
@@ -232,6 +232,22 @@ def test_run_with_preserve_fds(podman_client: DockerClient):
     ):
         assert os.read(read_fd, 7) == b"foobar\n"
 
+def test_run_with_timezone(podman_client: DockerClient):
+    output = podman_client.container.run(
+        "ubuntu", ["date", "+'%Z'"], tz="local", remove=True
+    )
+    # Check the timezone in the container matches that in the host.
+    assert output == datetime.now().astimezone().strftime("'%Z'")
+
+def test_create_start_with_timezone(podman_client: DockerClient):
+    with podman_client.container.create(
+        "ubuntu", ["sleep", "infinity"], tz="GMT", stop_timeout=1
+    ) as container:
+        container.start()
+        output = podman_client.execute(container, ["date", "+'%Z'"])
+        # Check the timezone in the container matches what was specified
+        assert output == "'GMT'"
+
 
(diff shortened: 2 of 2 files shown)


python_on_whales/components/container/cli_wrapper.py
```diff
@@ -626,6 +626,7 @@ def create(
         systemd: Optional[Union[bool, Literal["always"]]] = None,
         tmpfs: List[ValidPath] = [],
         tty: bool = False,
+        tz: Optional[str] = None,
```
legaul-cisco, line 629: Missing update to docstring. The PR description says "as supported by podman and docker", but is it supported by docker? I can't see it at https://docs.docker.com/reference/cli/docker/container/run/#options

LewisGaul, line 629: Missing update to docstring. The PR description says "as supported by podman and docker", but is it supported by docker? I can't see it at https://docs.docker.com/reference/cli/docker/container/run/#options

LewisGaul: In fact you've implemented the tests to only run with podman, was that intentional? I think just need to say in the docstring that it's podman only, and update the PR description for clarity.

tjd78: I was mistaken in my PR description, so I'll make this clear.

tests/python_on_whales/components/test_container.py
```diff
@@ -232,6 +232,22 @@ def test_run_with_preserve_fds(podman_client: DockerClient):
...
+def test_run_with_timezone(podman_client: DockerClient):
+    output = podman_client.container.run(
+        "ubuntu", ["date", "+'%Z'"], tz="local", remove=True
+    )
+    # Check the timezone in the container matches that in the host.
+    assert output == datetime.now().astimezone().strftime("'%Z'")
+
+def test_create_start_with_timezone(podman_client: DockerClient):
+    with podman_client.container.create(
+        "ubuntu", ["sleep", "infinity"], tz="GMT", stop_timeout=1
```
LewisGaul, line 244: I'm not sure why you have `stop_timeout=1` here - if you actually wanted to remove the timeout you'd want `stop_timeout=0`, but I wouldn't have thought it'd be needed with `sleep` which should get stopped immediately by `SIGTERM`?

tests/python_on_whales/components/test_container.py
```diff
@@ -232,6 +232,22 @@ def test_run_with_preserve_fds(podman_client: DockerClient):
...
+    ) as container:
+        container.start()
+        output = podman_client.execute(container, ["date", "+'%Z'"])
+        # Check the timezone in the container matches what was specified
+        assert output == "'GMT'"
```
LewisGaul, line 249: I take it GMT was chosen as being different to the default in the `ubuntu` image, which seems to be UTC? Maybe worth a comment

tests/python_on_whales/components/test_container.py
```diff
@@ -232,6 +232,22 @@ def test_run_with_preserve_fds(podman_client: DockerClient):
...
+    )
+    # Check the timezone in the container matches that in the host.
+    assert output == datetime.now().astimezone().strftime("'%Z'")
+
+def test_create_start_with_timezone(podman_client: DockerClient):
+    with podman_client.container.create(
+        "ubuntu", ["sleep", "infinity"], tz="GMT", stop_timeout=1
+    ) as container:
+        container.start()
+        output = podman_client.execute(container, ["date", "+'%Z'"])
```
LewisGaul, line 247: Why do you have quotes around the `%Z`?
```suggestion
        output = podman_client.execute(container, ["date", "+%Z"])
```

tests/python_on_whales/components/test_container.py
```diff
@@ -232,6 +232,22 @@ def test_run_with_preserve_fds(podman_client: DockerClient):
...
         assert os.read(read_fd, 7) == b"foobar\n"
 
+def test_run_with_timezone(podman_client: DockerClient):
+    output = podman_client.container.run(
+        "ubuntu", ["date", "+'%Z'"], tz="local", remove=True
+    )
+    # Check the timezone in the container matches that in the host.
+    assert output == datetime.now().astimezone().strftime("'%Z'")
+
+def test_create_start_with_timezone(podman_client: DockerClient):
```
LewisGaul, line 242: There should be two blank lines here which is causing the linting job to fail - run black to fix

tests/python_on_whales/components/test_container.py
tjd78: was copied from another test case I don't think it is needed.

tests/python_on_whales/components/test_container.py
tjd78: To match the format of the check. No particular reason - it doesn't really matter what the format is exactly so long as they match but I can get rid of the superfluous characters.

python_on_whales/components/container/cli_wrapper.py
```diff
@@ -1549,6 +1552,9 @@ def run(
                 https://docs.podman.io/en/latest/markdown/podman-run.1.html#systemd-true-false-always
             tty: Allocate a pseudo-TTY. Allow the process to access your terminal
                 to write on it.
+            tz: Set timezone in container, or `local` to match the host's timezone.
+                See `/usr/share/zoneinfo/` for valid timezones.
```
LewisGaul, line 1556: flake8 error: https://github.com/gabrieldemarmiesse/python-on-whales/actions/runs/8738375248/job/23980161755?pr=579
`./python_on_whales/components/container/cli_wrapper.py:1556:64: W291 trailing whitespace`

LewisGaul (approved):

LewisGaul:
@gabrieldemarmiesse did you want to have a look? Should be pretty uncontroversial :)

gabrieldemarmiesse (approved):
Looks good to me, we can merge this. I would be open to accept in the future a PR to accept [`datetime.timezone`](https://docs.python.org/3/library/datetime.html#datetime.timezone) as a valid argument for "tz". Of course str and None should always be accepted.

Thank you @LewisGaul  for driving the code review!

PR #583, opened 2024-05-01
Allow `podman exec` with `--interactive` and no `--tty`

jhc4318:
An interactive container does not need to have a TTY. There's a slight difference in behaviour that is currently a blocker for my use case.

```
>podman exec -i ctr_name /bin/bash -c 'hi' 1> 1.txt 2> 2.txt
>cat 1.txt
>cat 2.txt
/bin/bash: hi: command not found

>podman exec -it ctr_name /bin/bash -c 'hi' 1> 1.txt 2> 2.txt
>cat 1.txt 
/bin/bash: hi: command not found
>cat 2.txt 
```

I've added a testcase for the `--interactive` only case however stdout/stderr isn't captured by PoW in the `-it` case so no test has been added there (this also does not affect the existing behaviour there anyway).

### DIFF
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -904,13 +904,6 @@ class ContainerCLI(DockerCLICaller):
         full_cmd.add_args_list("--env", format_dict_for_cli(envs))
         full_cmd.add_args_list("--env-file", env_files)
 
-        # TODO: activate interactive and tty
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse (approved):
Many thanks for the pull request and sorry for the delay. This PR is short, has a clear explanation and a unit test. All the good stuff :) PRs like this make my day!

PR #587, opened 2024-05-11
Drop typing-extensions in recent Python versions

Dreamsorcerer:
It would be best to drop typing-extensions as a dependency when it's no longer needed. This removes the dependency on Python 3.11+.

### DIFF
--- a/python_on_whales/components/buildx/imagetools/models.py
+++ b/python_on_whales/components/buildx/imagetools/models.py
@@ -4,6 +4,7 @@ import sys
 from typing import List, Optional
 
 import pydantic
+
 if sys.version_info >= (3, 9):
     from typing import Annotated
 else:
--- a/python_on_whales/components/buildx/imagetools/models.py
+++ b/python_on_whales/components/buildx/imagetools/models.py
@@ -1,9 +1,13 @@
 from __future__ import annotations
 
+import sys
 from typing import List, Optional
 
 import pydantic
-from typing_extensions import Annotated
+if sys.version_info >= (3, 9):
+    from typing import Annotated
+else:
+    from typing_extensions import Annotated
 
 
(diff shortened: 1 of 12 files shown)


gabrieldemarmiesse:
Thank you for the pull request. I appreciate it and I see that in the short term, it will drop a dependency, which is always good.

I'm more worried about the long term. typing_extensions is extremely useful in general. It has no dependencies itself, so having the latest version isn't usually a problem. As python introduces new typing features, we'll likely use them. So we will need to add typing-extensions again. To avoid this back and forth, I believe that making typing-extensions a permanent dependencies of python-on-whales is reasonnable. It will make the adoption of future typing features much easier.

I'll close this PR, but if you believe there is something I'm missing, do not hesitate to continue this conversation :)

Dreamsorcerer:
I think people typically prefer to have less dependencies when they are not needed (this can also affect downstream decisions on packaging/supporting a package). I think anything past 3.11 are generally going to be nice-to-have features that can be utilised once upgrading to that version of Python, as opposed to important features that are needed for typing to actually work on the project. You'll notice that currently there is only 1 thing being used from typing-extensions (with 1 more being added in #580).

It's up to you, but I wouldn't anticipate any _requirement_ to use typing-extensions in future in order to correctly type the project (i.e. It should be easy to resist re-adding typing-extensions in future).

LewisGaul:
Personally I'm not too worried about the `typing-extensions` package being a dependency (as a user), and I can understand wanting to be free to use all the latest typing features (that's the whole point of the package, right? :) ). That's coming from a position where I definitely agree that I prefer to have less dependencies (see https://github.com/gabrieldemarmiesse/python-on-whales/issues/512#issuecomment-1859292421).

PR #589, opened 2024-05-14
Fix documentation and docstring issues

pcharmoille:
I still did not manage to serve the documentation locally but this is a start.


gabrieldemarmiesse (approved):
Many thanks for the fix! I actually noticed it just before reading your PR haha

PR #590, opened 2024-05-21
Allow `List[str]` in `DockerContainerListFilters['filters']` and `format_dict_for_cli()`

LooLzzz:
The docker cli allows for multiple `--filter` args to be passed, 
this PR aims to imitate that.

https://docs.docker.com/reference/cli/docker/container/ls/#filter

### DIFF
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -58,7 +58,7 @@ DockerContainerListFilters = TypedDict(
     {
         "id": str,
         "name": str,
-        "label": str,
+        "label": Union[str, List[str]],
         "exited": int,
         "status": Literal[
             "created", "restarting", "running", "removing", "paused", "exited", "dead"
--- a/python_on_whales/utils.py
+++ b/python_on_whales/utils.py
@@ -295,8 +295,21 @@ def stream_stdout_and_stderr(
         raise DockerException(full_cmd, exit_code, stderr=full_stderr)
 
 
-def format_dict_for_cli(dictionary: Dict[str, str], separator="="):
-    return [f"{key}{separator}{value}" for key, value in dictionary.items()]
+def format_dict_for_cli(dictionary: Dict[str, Union[str, List[str]]], separator="=") -> List[str]:
+    result_list = []
+
+    for key, value in dictionary.items():
+        if isinstance(value, str):
+            result_list.append(
+                f"{key}{separator}{value}"
+            )
+        elif isinstance(value, (list, tuple)):
+            result_list.extend(
+                f"{key}{separator}{v}"
+                for v in value
+            )
+
+    return result_list
 
 
 def read_env_file(env_file: Path) -> Dict[str, str]:


python_on_whales/components/container/cli_wrapper.py
```diff
@@ -58,7 +58,7 @@
     {
         "id": str,
         "name": str,
-        "label": str,
+        "label": Union[str, List[str]],
```
LewisGaul, line 61: The problem I have with this is it special-cases `label`, when actually it's presumably valid to specify any of the filter types more than once? It seems like conceptually what we want is to allow specifying the same key twice, but that's not possible with normal dicts in python of course.

Any thoughts @gabrieldemarmiesse on how to resolve this? One option would be to accept a list of 2-tuples instead of a dict, i.e. something like `docker.container.list(filters: Sequence[DockerContainerListFilters])` where `DockerContainerListFilters = tuple[Literal["id", str]] | tuple[Literal["name"], str] | ...`.

LewisGaul (commented):
Hey @LooLzzz thanks for the contribution :) you've got some failures in the CI, but there's also a comment I had about the approach which it'd be good to get thoughts from @gabrieldemarmiesse on.

python_on_whales/components/container/cli_wrapper.py
LewisGaul: What I don't like about that is it's not truly mirroring the CLI (you don't specify `--filter label=foo,bar`), so it feels like sacrificing the quality of the API for minor backwards compatibility (could easily enough continue to support the old form until v1.0 if we wanted). Perhaps not all fields would make sense to be specified more than once, is that something we'd need to think about in your suggested approach?

A couple of other notes on this API:
- Why is there a `name` field, should that be `reference`? (https://docs.docker.com/reference/cli/docker/image/ls/#filter, https://docs.podman.io/en/stable/markdown/podman-images.1.html#filter-f-filter)
- Podman also supports `--filter label!=foo`, which the current API does not support

gabrieldemarmiesse: That would be worth a redesign from scratch possibly... As we're pre-1.0 I believe we can afford it

LooLzzz: Maybe chained calls would suite this API better?  
[Many ORMs](https://docs.djangoproject.com/en/5.0/topics/db/queries/#chaining-filters) are using this style of API

```py
docker.container.list()
                .filter({'label': foo})
                .filter({'label': bar})
```

LewisGaul: This approach has downsides that make me hesitant:
* implementation complexity
* cognitive overhead caused by the fact that under certain APIs the actual command gets run at expected times

I still think we should go for `docker.container.list(filters=[("label", "foo"), ("label", "bar=1"), ("label!", "foo=0"), ("status", "created"), ("status", "running"), ("name", "prefix-*")])` (note that the CLI is a bit confusing here in that label filters are ANDed together whereas status filters are ORed together).

LewisGaul:
Tracking the core of the issue at https://github.com/gabrieldemarmiesse/python-on-whales/issues/629

LewisGaul:
Closing in favour of https://github.com/gabrieldemarmiesse/python-on-whales/pull/635 - thanks for bringing this issue up, it'll be good to have this solved properly.

Issue #591, opened 2024-05-22
Implement `DockerClient.version()`


LewisGaul:
This would be useful for skipping certain tests that cover features not available on older versions, as well as perhaps being needed in the main code at some point if we care about supporting more than just the latest version of docker/podman. Users may also find this useful.

Note the API already exists, it just raises `NotImplementedError` - just creating this issue to track/discuss it and have something to link to from other issues :)

LewisGaul:
Would be good to discuss how we want to solve this.

Docker/podman have two version commands: `docker --version` and `docker version`. The former is the simple client version, returning something like "Docker version 20.10.23, build 7155243". The latter connects to the server/engine side (which may be on a remote host) so can fail (meaning it can be used as a basic connectivity check) and provides more information. I think `DockerClient.version()` should correspond to the latter to provide that full information (server-side version is likely more relevant in general, and the connection check can be useful). [Aside: awkwardly, when you want to check whether a feature is supported you might actually have to care about the minimum of the client and server versions...]

Examples:
* With local docker server connection
```
root@ubuntu:~# docker version
Client: Docker Engine - Community
 Version:           20.10.23
 API version:       1.41
 Go version:        go1.18.10
 Git commit:        7155243
 Built:             Thu Jan 19 17:36:25 2023
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.23
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.10
  Git commit:       6051f14
  Built:            Thu Jan 19 17:34:14 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.21
  GitCommit:        3dce8eb055cbb6872793272b4f20ed16117344f8
 runc:
  Version:          1.1.7
  GitCommit:        v1.1.7-0-g860f061
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```
* With remote docker connection, using JSON format
```
   > docker version -f '{{json .}}' | jq
{
  "Client": {
    "Platform": {
      "Name": "Docker Engine - Community"
    },
    "Version": "20.10.17",
    "ApiVersion": "1.41",
    "DefaultAPIVersion": "1.41",
    "GitCommit": "100c701",
    "GoVersion": "go1.17.11",
    "Os": "linux",
    "Arch": "amd64",
[...]

PR #593, opened 2024-05-22
Add support for 'docker version' and 'podman version'

LewisGaul:
Fixes https://github.com/gabrieldemarmiesse/python-on-whales/issues/591

- [x] Resolve what to do about the added requirement for pydantic v2
   - Making it work with pydantic v1 for now (although podman's `APIVersion` and `BuiltTime` fields won't be recognised with pydantic v1)
   - Raised separate issue for removing pydantic v1 support: https://github.com/gabrieldemarmiesse/python-on-whales/issues/594
- [x] Add tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,3 +37,4 @@ markers = [
     "podman: marks tests as needing a podman engine to run"
 ]
 xfail_strict = true
+pythonpath = ["./"]
--- a/python_on_whales/__init__.py
+++ b/python_on_whales/__init__.py
@@ -17,7 +17,7 @@ from .components.stack.cli_wrapper import Stack
 from .components.system.cli_wrapper import SystemInfo
 from .components.task.cli_wrapper import Task
 from .components.volume.cli_wrapper import Volume
-from .docker_client import DockerClient
+from .docker_client import DockerClient, Version
 from .exceptions import DockerException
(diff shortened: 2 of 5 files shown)


gabrieldemarmiesse:
I would be happy to drop support for pydantic V1 when it's end of life, so in one month. If this PR is not a pressing matter, we can do it at this point in time :)

pyproject.toml
```diff
@@ -37,3 +37,4 @@ markers = [
     "podman: marks tests as needing a podman engine to run"
 ]
 xfail_strict = true
+pythonpath = ["./"]
```
LewisGaul, line 40: This makes just `pytest` work from the project root, rather than needing `python3 -m pytest`

gabrieldemarmiesse (commented):
Looks good as usual! We can merge when pydantic v1 is EOL

gabrieldemarmiesse (approved):

LewisGaul:
> Looks good as usual! We can merge when pydantic v1 is EOL

To be clear, I updated it to be compatible with pydantic v1 so that we don't need to wait.

gabrieldemarmiesse:
Ho my bad, I went over it a bit too quickly since I already read the diff once. Let's merge it now :)

Issue #584, opened 2024-05-01
Avoid using `List` and `Dict` for function argument types


LewisGaul:
For example:
```python
ValidImage = Union[str, Image]

class ImageCLI(DockerCLICaller):
    def remove(
        self,
        x: Union[ValidImage, List[ValidImage]],
        force: bool = False,
        prune: bool = True,
    ): ...
```

If you call this as so:
```python
from python_on_whales import docker
my_tags = ["foo", "bar"]
docker.image.remove(my_tags)
```

Then running mypy gives:
```
tmp.py:4: error: Argument 1 to "remove" of "ImageCLI" has incompatible type "list[str]"; expected "str | Image | list[str | Image]"  [arg-type]
tmp.py:4: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
tmp.py:4: note: Consider using "Sequence" instead, which is covariant
```

The reason that mypy is complaining is that the API accepts a `List[str | Image]`, i.e. a list that can hold strings or images, but not a list of only strings. Lists are mutable, so with the function could modify or add elements - and (according to the type signature) it's allowed to add elements that are of type `Image`, which would violate the inferred type of `my_tags`.

One solution on the caller side is to annotate `my_tags: List[str | Image]`, but this could require importing `Image` and generally just shouldn't be necessary. PoW isn't intending to mutate the list so this should be indicated using `Sequence` or `Iterable` or similar.

gabrieldemarmiesse:
I don't mind typing with something more generic than a list. After all, we often just iterate on the arguments. 

As long as it doesn't complexify the codebase too much, I'd be happy to accept a PR in this direction. 

A small PR would be great so that we can have an example to illustrate the change and its consequences.

LewisGaul:
It's not really a problem of being more generic, e.g. allowing sets/tuples (although that would be nice too). The problem is that if you annotate with `List[str | Image]` then the input must also be a `List[str | Image]` - passing a `List[str]` is invalid simply because the argument (a list) is a mutable type, even though PoW intends this to be valid since it won't mutate the argument.

Unfortunately if we switch to a more generic type we'll also have to change `isinstance(x, list)` checks...

gabrieldemarmiesse:
I don't see many other solution than using a more generic type. And yes it will change the isinstance() calls but I believe that's a reasonnable amount of work and that the benefits are worth it.

PR #595, opened 2024-05-25
Set up command construction to support generic iterables and mappings

LewisGaul:
First step in working on https://github.com/gabrieldemarmiesse/python-on-whales/issues/584.

- `Command.add_args_list()` previously had a misleading name and type annotation - it is used for accepting either a list or a single item. Renamed to `Command.add_args_iterable_or_single()`.
- Added `Command.add_args_iterable()` for the case you know you're dealing with an iterable (more efficient than always creating a new list).
- Updated `utils.to_list()` to support generic iterables (still always returns a list, as per the name - my plan is to phase out its use where it's not needed).
- Updated `utils.format_dict_for_cli()` to support generic mappings.
- Added `Command.add_args_mapping()` as a convenience that wraps `utils.format_dict_for_cli()`, which will allow tidying up code that constructs commands.

Next step is to tidy up the components' code that constructs commands, allowing arguments to be `Iterable` and `Mapping` in place of `List` and `Dict`. With these changes it should be fairly straightforward, simply calling the correct method on the `Command` class.

Here's an example of what upcoming changes look like: https://github.com/LewisGaul/python-on-whales/compare/command-construction...command-construction-image

### DIFF
--- a/python_on_whales/client_config.py
+++ b/python_on_whales/client_config.py
@@ -5,7 +5,7 @@ import warnings
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta
 from pathlib import Path
-from typing import Any, Dict, List, Literal, Optional, Union
+from typing import Any, Dict, Iterable, List, Literal, Optional, Union
 
 import pydantic
 
@@ -37,8 +37,10 @@ class Command(list):
         if value:
             self.append(name)
 
-    def add_args_list(self, arg_name: str, list_values: list):
-        for value in to_list(list_values):
+    def add_args_iterable_or_single(
+        self, arg_name: str, iterable_or_single: Union[Iterable[Any], Any]
+    ):
+        for value in to_list(iterable_or_single):
             self.extend([arg_name, value])
 
     def __add__(self, other) -> "Command":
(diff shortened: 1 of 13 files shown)


python_on_whales/client_config.py
```diff
@@ -37,10 +37,23 @@ def add_flag(self, name: str, value: bool):
         if value:
             self.append(name)
 
-    def add_args_list(self, arg_name: str, list_values: list):
-        for value in to_list(list_values):
+    def add_args_iterable(self, arg_name: str, values: Iterable[Any]):
+        for value in values:
             self.extend([arg_name, value])
```
gabrieldemarmiesse, line 42: I don't think this is needed. It is indeed faster than the other function and more efficient, but as long as the performance drop is not noticeable, let's not add more code to this codebase.

I had measure this a while algo, and most of the time spent in python-on-whales is actually spent calling subprocess. I double such an optimisation will have an impact

gabrieldemarmiesse (changes requested):
Looks good overall. We can drop the speed optimisation. I don't think anyone is going to notice.

I usually refuse code that is added "to make things faster" unless there is a benchmark showing that the speed increase is noticeable.

python_on_whales/client_config.py
LewisGaul: Sure, but it's not about performance, it's a correctness thing. I think this kind of improvement is highly valuable, it more clearly (and *correctly*) conveys the types being passed around rather than a general "best-effort" type of approach. If a single value is accidentally passed to this method it would cause an error, which is desirable. This kind of thing would help avoid problems like https://github.com/gabrieldemarmiesse/python-on-whales/issues/466.

gabrieldemarmiesse: I see what you mean. I would believe that would be necessary for public functions, but since it's not public, I see it as less useful. But anyway, I don't feel that strongly about this, if you see its value when doing your refactoring, it's good enough for me. I'll approve it :)

gabrieldemarmiesse (approved):

PR #599, opened 2024-06-01
Update pod component to use generic Iterable and Mapping types

LewisGaul:
Follows https://github.com/gabrieldemarmiesse/python-on-whales/pull/595, working on https://github.com/gabrieldemarmiesse/python-on-whales/issues/584.


LewisGaul:
> I notice I actually used `Sequence` in some of the other methods... I'm now questioning whether `Iterable` or `Sequence` (or something else?) is preferably here, especially considering a string itself is a valid `Iterable[str]`...

The "strings are iterables of strings" issue is tracked at https://github.com/python/typing/issues/256. The pondering about `Iterable` vs. `Sequence` isn't affected though, since strings are also sequences of strings. I think `Iterable` should be preferable, to allow the user to pass generators (e.g. `docker.volume.inspect((x for x in volume_names if x.startswith("foo"))`).

gabrieldemarmiesse (approved):
Looks good! Thanks for the patch!

PR #600, opened 2024-06-02
Add bootstrap cli flag for create/inspect

fizzgig1888:
Hello,
The purpose of this PR is to add the hability to use the --bootstrap optional flag for the following commands :
`$ docker buildx create [OPTIONS] [CONTEXT|ENDPOINT]`
`$ docker buildx inspect [NAME]`

### DIFF
--- a/python_on_whales/components/buildx/cli_wrapper.py
+++ b/python_on_whales/components/buildx/cli_wrapper.py
@@ -507,7 +507,7 @@ class BuildxCLI(DockerCLICaller):
     def inspect(
         self,
         x: Optional[str] = None,
-        bootstrap: Optional[bool] = False,
+        bootstrap: bool = False,
     ) -> Builder:
         """Returns a builder instance from the name.
 
--- a/python_on_whales/components/buildx/cli_wrapper.py
+++ b/python_on_whales/components/buildx/cli_wrapper.py
@@ -454,6 +454,7 @@ class BuildxCLI(DockerCLICaller):
     def create(
         self,
         context_or_endpoint: Optional[str] = None,
+        bootstrap: bool = False,
         buildkitd_flags: Optional[str] = None,
         config: Optional[ValidPath] = None,
         platforms: Optional[List[str]] = None,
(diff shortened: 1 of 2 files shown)


python_on_whales/components/buildx/cli_wrapper.py
```diff
@@ -501,16 +504,25 @@ def disk_usage(self):
...
         Parameters:
             x: If `None` (the default), returns the current builder. If a string is provided,
                 the builder that has this name is returned.
-
+            bootstrap: If set to True, ensure builder has booted before inspecting.
+                
         # Returns
             A `python_on_whales.Builder` object.
         """
+        if bootstrap:
```
fizzgig1888, line 522: Note that running the `docker buildx inspect --boostrap [NAME] ` command is necessary, when the bootstrap flag is set to True, as the command modifies the targeted builder state.

gabrieldemarmiesse: That seems reasonable

python_on_whales/components/buildx/cli_wrapper.py
```diff
@@ -501,16 +504,27 @@ def disk_usage(self):
         """Not yet implemented"""
         raise NotImplementedError
 
-    def inspect(self, x: Optional[str] = None) -> Builder:
+    def inspect(
+        self,
+        x: Optional[str] = None,
+        bootstrap: Optional[bool] = False,
```
gabrieldemarmiesse, line 510:
```suggestion
        bootstrap: bool = False,
```
`None` is not an acceptable value, as such, it's not an optional.

gabrieldemarmiesse (changes requested):
Thank you for the pull request, it's a cool feature! Also sorry for the delay in reviewing. Could you just add a small unit test? After that it should be good to merge. Thank you!

python_on_whales/components/buildx/cli_wrapper.py
fizzgig1888: Fixed !

fizzgig1888:
Hello, and many thanks for the review and advices. Don't worry for the delay, I wasn't quick either.

Two unit tests were added : test_buildx_inspect_bootstrap and test_buildx_create_bootstrap, for checking the builder bootstrap effectiveness, as you requested :)

gabrieldemarmiesse (approved):
Looks good! Many thanks for this feature!

Issue #606, opened 2024-07-08
UnicodeDecodeError getting container output


LewisGaul:
This is seen when calling `DockerClient.container.run()` or `DockerClient.container.logs()` if the container console contains output that is not UTF-8 encoded. This can be reproduced by outputting a character in another encoding as shown below.

Related to https://github.com/gabrieldemarmiesse/python-on-whales/issues/370. Perhaps `DockerClient.container.logs()` should take an `encoding` parameter rather than assuming utf-8.

```
>>> import python_on_whales as pow
>>> podman = pow.DockerClient(client_call=["podman"])
>>> bash_cmd = r"echo -n $'\xfe'"
>>> full_cmd = ["bash", "-c", bash_cmd]
>>> ctr = podman.run("ubuntu", full_cmd)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../python_on_whales/components/container/cli_wrapper.py", line 1764, in run
    return run(full_cmd, tty=tty, capture_stderr=False, pass_fds=pass_fds)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../python_on_whales/utils.py", line 215, in run
    return post_process_stream(completed_process.stdout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../python_on_whales/utils.py", line 221, in post_process_stream
    stream = stream.decode()
             ^^^^^^^^^^^^^^^
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xfe in position 0: invalid start byte
```

PR #607, opened 2024-07-10
Replace unrecognised unicode characters rather than raising an exception

LewisGaul:
Fixes https://github.com/gabrieldemarmiesse/python-on-whales/issues/606

Note that this still isn't ideal - if someone has a container expected to have output in an encoding other than unicode then special characters will all be replaced... It might be good to support the user passing `encoding` and `errors` through (much like [`subprocess.run()`](https://docs.python.org/3/library/subprocess.html#subprocess.run)), but for now this seems like an improvement at least.

### DIFF
--- a/python_on_whales/utils.py
+++ b/python_on_whales/utils.py
@@ -237,10 +237,10 @@ def run(
         return post_process_stream(completed_process.stdout)
 
 
-def post_process_stream(stream: Optional[bytes]):
+def post_process_stream(stream: Optional[bytes]) -> str:
     if stream is None:
         return ""
-    stream = stream.decode()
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse (approved):
I agree this looks okay, and I also agree that users should be able to pass the encoding they want. Also streaming the output to get raw bytes is also an option I believe and it already works with python-on-whales: https://gabrieldemarmiesse.github.io/python-on-whales/user_guide/docker_run/#stream-the-output

LewisGaul:
I raised https://github.com/gabrieldemarmiesse/python-on-whales/issues/608 to track the enhancement of allowing the user to specify encoding.

PR #615, opened 2024-08-01
Add support for 'podman container init'

LewisGaul:
I haven't added the `DockerClient.init()` alias although `podman init` is supported, because it's not supported by docker... Should I add it?


gabrieldemarmiesse (approved):
Really nice PR. About your question, since docker users are the vast majority of Python-on-whales users, I would prefer to skip the alias for now, as this can degrade the DX. 

Notably, we don't throw any exception currently we someone is using a function unsuported for their container engine. So we will want to avoid that as much as possible.

Issue #553, opened 2024-02-14
Docker compose support for multiple env files (--env-file, compose_env_file)


marcinhlybin:
The `docker-compose` command includes the `--env-file` option, which accepts multiple files by allowing the option to be repeated. This is indicated in the command help as `stringArray`, meaning it can take an array of strings, each representing a path to an environment file.

To enhance functionality, a proposed update introduces the `compose_env_files` setting, which accepts a `List[ValidPath]`. This addition enables specifying multiple environment files directly in the configuration and it uses the same naming convention as for other options. 

However, removing the existing `compose_env_fil`e setting would lead **to breaking changes due to loss of backward compatibility**.

A more user-friendly approach might be to retain `compose_env_file` for compatibility while expanding its capabilities to accept either a single path or a list of paths. This dual support ensures backward compatibility and introduces the flexibility of specifying multiple environment files, either through the new `compose_env_files` setting or by extending the functionality of the existing `compose_env_file` option.

```
diff --git a/python_on_whales/client_config.py b/python_on_whales/client_config.py
index 207961c..fb8b045 100644
--- a/python_on_whales/client_config.py
+++ b/python_on_whales/client_config.py
@@ -61,6 +61,7 @@ class ClientConfig:
     compose_files: List[ValidPath] = field(default_factory=list)
     compose_profiles: List[str] = field(default_factory=list)
     compose_env_file: Optional[ValidPath] = None
+    compose_env_files: List[ValidPath] = field(default_factory=list)
     compose_project_name: Optional[str] = None
     compose_project_directory: Optional[ValidPath] = None
     compose_compatibility: Optional[bool] = None
@@ -147,6 +148,7 @@ class ClientConfig:
         base_cmd.add_args_list("--file", self.compose_files)
         base_cmd.add_args_list("--profile", self.compose_profiles)
         base_cmd.add_simple_arg("--env-file", self.compose_env_file)
+        base_cmd.add_args_list("--env-file", self.compose_env_files)
         base_cmd.add_simple_arg("--project-name", self.compose_project_name)
         base_cmd.add_simple_arg("--project-directory", self.compose_project_directory)
         base_cmd.add_flag("--compatibility", self.compose_compatibility)
```

gabrieldemarmiesse:
I would be happy with such a feature indeed, thanks for opening this issue. PR welcome!

einarwar:
[...]

PR #616, opened 2024-08-01
Allow for multiple env-files

einarwar:
Sorry for the mess with multiple PRs, not used to making PR's from forks :)

Closes https://github.com/gabrieldemarmiesse/python-on-whales/issues/553

Keeps backwards compatibility with the compose_env_file parameter, but advises the user to use the new compose_env_files parameter instead. If both parameters are given, the compose_env_files will take presedence.

Added tests to verify behavior


tests/python_on_whales/test_client_config.py
```diff
@@ -36,3 +37,53 @@ def test_pretty_exception_message_and_report(mocker):
...
+        index
+        for index, item in enumerate(client_config.docker_compose_cmd)
+        if item == "--env-file"
+    ]
+    assert len(indices) == len(compose_env_files)
+    for index, env_file in zip(indices, compose_env_files):
+        assert client_config.docker_compose_cmd[index + 1] == env_file
+
+
+def test_compose_env_files_and_env_file():
```
gabrieldemarmiesse, line 75: The tests are good, keep them :) I'd also like a test which actually runs the docker compose command and checks that the env files are loaded correctly. For example with `docker.compose.config()`. If for example, you made a typo in `--env-file` in the unit test and the main implementation, your tests can't currently detect it. Running the true command in test would detect such an error.

gabrieldemarmiesse (approved):
The implementation looks good! Many thanks! I think we just need one additional test to be confident about the implementation

gabrieldemarmiesse (changes requested):
Sorry, I meant to request changes

tests/python_on_whales/test_client_config.py
einarwar: Added some more tests now that actually run the command by using the equivalent of `docker compose run`. Some temporary env-files are made, and the containers are then inspected to verify that the actual environment variables set in the container are as expected. Pretty verbose, and I am sure it can be done simpler, but let me know if this is what you wanted :)

einarwar: Aha, yeah that is a lot simpler. Updated now :)

tests/python_on_whales/components/dummy_compose.yml
```diff
@@ -24,8 +24,8 @@ services:
     image: alpine:latest
     command: sleep infinity
     environment:
-     - DD_API_KEY=__your_datadog_api_key_here__
-     - POSTGRES_HOST_AUTH_METHOD=trust
+      - DD_API_KEY=__your_datadog_api_key_here__
```
gabrieldemarmiesse, line 27: I'll let those formatting changes slide, but I'll ask that in the future you avoid adding formatting changes when they are not necessary. I think your IDE did it automatically, maybe you should disable auto-formatting yaml in this project.

gabrieldemarmiesse (approved):
Looks good! Thanks!

tests/python_on_whales/components/dummy_compose.yml
einarwar: Oh sorry, i did not notice that. Will disable for future contributions

PR #617, opened 2024-08-01
Replace isort, black and flake8 with ruff

einarwar:
[Ruff](https://docs.astral.sh/ruff/) has emerged as the most popular tool for linting, and covers the rules from flake8 among others. It also serves as an (almost) drop in replacement for black and isort.

This PR replaces flake8, isort and black with ruff. 
* There are some minor differences is how black and ruff formats the files, as seen in the files formatted in this PR
* Ruff with `F`, `E` and `W` covers the rules by the existing flake8 according to [these docs](https://docs.astral.sh/ruff/faq/#how-does-ruff-compare-to-flake8)
* Updated pre-commit and github actions to also use Ruff


LewisGaul:
> [Ruff](https://docs.astral.sh/ruff/) has emerged as the most popular tool for linting, and covers the rules from flake8 among others. It also serves as an (almost) drop in replacement for black and isort.

Ruff is great, especially for linting. I have a slight concern about switching over from black/isort given the [minor] differences, meaning that IDEs then need to be reconfigured to *not* use black/isort and use ruff instead, and I'm not sure ruff is currently used more widely than black for formatting. Otherwise I'm in favour, with the biggest win being `ruff check --fix`.

gabrieldemarmiesse:
I'm in favor of using ruff for everything (isort, flake8, and black replacement). While it is true that it is not as used as black and isort, there is a big win in developer experience. For those who don't wish to use the pre-commit for example. It's true that people will have to change the formatting in their IDE for those who use it, but if the pre-commit is installed anyway, the formatting will be fixed before commiting. It was long overdue that the Python ecosystem get a rust-like developer experience with tooling. 

A question I have. Will `ruff check --fix` run in the pre-commit? That would be great for the DX

einarwar:
> A question I have. Will `ruff check --fix` run in the pre-commit? That would be great for the DX

I think so. Will update :)

gabrieldemarmiesse (approved):
Looks good, thanks!

PR #628, opened 2024-09-01
Fix "Render docs" CI job

LewisGaul:
Hopefully fixing two issues that have started being hit in the "Render docs" CI job:
* `podman pod create` failing with "initializing source docker://k8s.gcr.io/pause:3.5: Requesting bear token: invalid status code from registry 404 (Not Found)", reported at https://github.com/gabrieldemarmiesse/python-on-whales/pull/625#issuecomment-2309580047 and seen in other PRs too.
  * Fixed by updating podman in the CI, since doc generation involves running podman commands.
* `ModuleNotFoundError: No module named 'griffe.collections'` reported at https://github.com/gabrieldemarmiesse/python-on-whales/pull/625#issuecomment-2322094393 and seen in other PRs too.
  * Fixed by updating doc package dependencies and pinning a couple more indirect dependencies (I think the lack of pinning `griffe` while pinning to an older version of `mkdocstrings-python` was the problem...).

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -18,8 +18,9 @@ jobs:
       - name: prepare
         run: |
           ./scripts/ci-podman-update.sh
+          pip install -U pip wheel
           pip install -r ./requirements.txt -r docs/requirements.txt -e .
-          python docs/autogen.py
+          ./docs/autogen.py
           cp -r docs/generated_sources ./generated_sources
           cp docs/mkdocs.yml ./
       - name: Render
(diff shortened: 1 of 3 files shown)


LewisGaul:
@gabrieldemarmiesse could you check the doc generation still works ok? I can't run it on my laptop since I don't have docker swarm set up. There are a bunch of warnings in the CI output about dead anchor links...

gabrieldemarmiesse:
Thanks, I tested locally your branch (I usually disable the pods docs since I don't have podman locally) and it worked fine. Let's merge this. I still get a lot of warnings, which I believe we should fix at some point. Though it's not urgent.

PR #632, opened 2024-09-02
Add "network exists" command

### LINKED ISSUE #618: Missing support for 'podman network exists'
Docker seems not to support this (at least on older versions), but podman does.

eclark0426:
Podman has support for a "network exists" command which checks if a network exists. This adds support for that command, although it does so by way of network inspect for compatibility with Docker, which does not have a standalone `network exists` command.

Closes #618


python_on_whales/components/network/cli_wrapper.py
```diff
@@ -209,6 +210,26 @@ def disconnect(
...
 
+    def exists(
+        self,
+        network: str,
+    ) -> bool:
+        """Check if a network exists
+
+        Although there is an actual exists command for Podman,
+        it is instead implemented using inspect for compatibility
+        with Docker, which does not support it as a standalone command.
```
LewisGaul, line 221: I think this belongs as a comment rather than a docstring - the caller shouldn't care how it's implemented

python_on_whales/components/network/cli_wrapper.py
```diff
@@ -209,6 +210,26 @@ def disconnect(
         full_cmd += [network, container]
         run(full_cmd)
 
+    def exists(
+        self,
+        network: str,
```
LewisGaul, line 215: Should support passing a `Network` object too, i.e. this should be `ValidNetwork` as with other methods such as `disconnect()` above.

python_on_whales/components/network/cli_wrapper.py
eclark0426: Seems like this same comment applies to other objects with an `exists()` method, and those don't have any similar info on the documentation, so I'll just remove it.

python_on_whales/components/network/cli_wrapper.py
eclark0426: Updated, along with making sure the test covers that use case.

python_on_whales/components/network/cli_wrapper.py
```diff
@@ -209,14 +217,30 @@ def disconnect(
...
+        try:
+            self.inspect(network)
+        except NoSuchNetwork:
+            return False
+        else:
+            return True
+
     @overload
-    def inspect(self, x: str) -> Network: ...
+    def inspect(self, x: ValidNetwork) -> Network: ...
```
LewisGaul, line 237: `inspect()` doesn't make a lot of sense to take a `Network` object, since it's used for *getting* a `Network` object. Can you instead make `exists()` do `self.inspect(str(network))`? (or check how his is handled for other object types such as `Container`, `Pod`, ...)

LewisGaul:
Thanks for doing this enhancement :) Currently all PRs are blocked on the docs CI job being broken, which I believe I have fixed in https://github.com/gabrieldemarmiesse/python-on-whales/pull/628 but this needs confirming.

python_on_whales/components/network/cli_wrapper.py
eclark0426: It seems like things are slightly inconsistent on whether or not `inspect()` takes `ValidSomeObject` or just `str`. This change was based on `ContainerCLI.inspect()`, which takes `ValidContainer` as an argument. `PodCLI.inspect()` similarly takes `ValidPod`. However, `VolumeCLI.inspect()` and `ServiceCLI.inspect()` only take a `str`.

Given the inconsistency, sounds like `str` would be the preferred approach, yes?

LewisGaul: Oh dear 🥲 I don't mind - the main thing I noticed is you didn't quite update the type annotations correctly in adding the support. I'd leave it as just `str` here to keep the scope of the changeset contained.

eclark0426: Moved back to `str` and used `self.id` in `Network.exists()`.

You'll have to excuse any errors in my type changes, I'm not all that familiar with the codebase. Been using it a lot the last several months, but only recently dived into the actual code. For my own understanding, what was incorrect with how I'd updated the type annotations?

LewisGaul: > For my own understanding, what was incorrect with how I'd updated the type annotations?

In [this commit](https://github.com/gabrieldemarmiesse/python-on-whales/commit/8d27a80ba7ade1ec7b9e66346f7cad0b90470d2d) you had
```python
    @overload
    def inspect(self, x: ValidNetwork) -> Network: ...

    @overload
    def inspect(self, x: List[str]) -> List[Network]: ...

    def inspect(self, x: Union[str, List[str]]) -> Union[Network, List[Network]]:
        if isinstance(x, str) or isinstance(x, Network):
            return Network(self.client_config, x)
        else:
            return [Network(self.client_config, reference) for reference in x]
```

* The `@overload` methods say that the two possible invocations are a single `ValidNetwork` (`str | Network`) or a list of `str`. This means there's not support for a list of `Network`, i.e. the second overload should've had `List[ValidNetwork]` (although `List` is a bad choice of type, see https://github.com/gabrieldemarmiesse/python-on-whales/issues/584).
* The actual method signature (without `@overload`) wasn't updated to support `Network` at all, so is inconsistent with the overloads.

If we wanted to support `Network` objects here I would make this:
```python
    @overload
    def inspect(self, x: ValidNetwork) -> Network: ...

    @overload
    def inspect(self, x: Iterable[ValidNetwork]) -> List[Network]: ...

    def inspect(self, x: Union[ValidNetwork, Iterable[ValidNetwork]]) -> Union[Network, List[Network]]:
        if isinstance(x, Iterable) and not isinstance(x, str):
            return [Network(self.client_config, str(net)) for net in x]
        else:
            return Network(self.client_config, str(x))
```

LewisGaul: To be clear, I still think the `exists()` method should support `Network` types:
```python
    def exists(self, network: ValidNetwork) -> bool:
        try:
            self.inspect(str(network))
        except NoSuchNetwork:
            return False
        else:
            return True
```

eclark0426: Got it. Overloads are one of things I haven't touched much with type annotations so it doesn't entirely surprise me that that would be where I went awry. Thanks for the clarification!

LewisGaul (approved):
Thanks again :)

Issue #575, opened 2024-04-02
Cache-installed docker doesn't include the compose plugin


faec:
Using python-on-whales to orchestrate CI machines, some commands produce the following failure:

```
failed on setup with "python_on_whales.exceptions.DockerException: The docker command executed was `/root/.cache/python-on-whales/docker-cli/20.10.5/docker compose [...]`.
It returned with code 125
The content of stdout is ''
The content of stderr is 'unknown flag: --file
```

Further inspection of the error output shows that docker doesn't recognize `compose` as a valid command. This seems to be because python-on-whales installed the base docker engine but not the compose plugin. Manually adding the plugin as in [this documentation](https://docs.docker.com/compose/install/linux/#install-the-plugin-manually) seems to fix the issue. This is only an issue on systems that don't already have docker (with compose) installed.

I'm preparing a PR to fix this case, but any extra context or advice would be welcome.

LewisGaul:
Hi @faec, one relevant point is that the plan is to remove the functionality of python-on-whales where it downloads the docker client, see https://github.com/gabrieldemarmiesse/python-on-whales/issues/512#issuecomment-1872322431.

faec:
@LewisGaul Ah, thank you! In that case maybe I'll work on fixing this in our CI initialization scripts instead

Issue #552, opened 2024-02-13
Slow import time


LewisGaul:
The `python_on_whales` package is a bit slow to import (noticeable for scripts when you just want to show help output):
```
   > python -V
Python 3.12.1

   > echo 'import python_on_whales' > /tmp/script.py

   > time python3 /tmp/script.py

real    0m0.629s
user    0m0.569s
sys     0m0.057s

   > python -m cProfile /tmp/script.py | head
         1438040 function calls (1220502 primitive calls) in 1.141 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       20    0.000    0.000    1.881    0.094 cli_wrapper.py:1(<module>)
       45    0.001    0.000    1.531    0.034 __init__.py:1(<module>)
    468/1    0.017    0.000    1.141    1.141 {built-in method builtins.exec}
        1    0.000    0.000    1.141    1.141 script.py:1(<module>)
   3502/1    0.005    0.000    1.141    1.141 <frozen importlib._bootstrap>:1349(_find_and_load)
```

LewisGaul:
Related to https://github.com/gabrieldemarmiesse/python-on-whales/issues/512 since importing these extra dependencies seems to be contributing around 25% of the time.

gabrieldemarmiesse:
Thanks for the bug report, I reproduced it with `tuna` and `-X importtime`, and here are the results I found:

* Importing python-on-whales takes around 480ms 
* Importing the download_cli module takes around 118ms, it will be fixed when #512 is fixed.
* The majority of the time left (80%) is used to create the pydantic models, see https://github.com/pydantic/pydantic/issues/6768

There is not much we can do about the pydantic models. We can provide solutions to users:

1) Users who are making CLI tools or other tools that are very sensitive to import times should use lazy import tools such as https://pypi.org/project/lazy-imports/ . I have used such tools in the past and found them very performant and easy to use.
2) Mojo is turning out to solve a bunch of issue that are present in python. For example, since it's a compiled language, the import times only impact the compilation speed and not the runtime speed. Once Mojo is mature enough, I would be very happy to make python-on-whales compatible. Users should be able to then make CLI tools that start very fast and are not impacted by import times by using the Mojo compiler.

Since there is not much we can do except fixing #512  and then much later (1+ year) provide a Mojo package, I would advise to close this issue.

LewisGaul:
[...]

PR #633, opened 2024-09-09
Remove download docker client functionality

LewisGaul:
Removes dependency on typer/tqdm/requests by dropping support for downloading the docker client.

Initially agreed at https://github.com/gabrieldemarmiesse/python-on-whales/issues/512#issuecomment-1872322431, this has been deprecated since https://github.com/gabrieldemarmiesse/python-on-whales/pull/577 (v0.71.0 released in April 2024), and agreed we can now remove at https://github.com/gabrieldemarmiesse/python-on-whales/issues/552#issuecomment-2331356820. If accepted, this supercedes https://github.com/gabrieldemarmiesse/python-on-whales/pull/513 as a resolution to https://github.com/gabrieldemarmiesse/python-on-whales/issues/512. We will also be able to close https://github.com/gabrieldemarmiesse/python-on-whales/issues/575.


LewisGaul:
@gabrieldemarmiesse are we ready to merge this do you think?

gabrieldemarmiesse (approved):
I'm deeply sorry for the late reply there. Thanks for the PR and the description. I believe we should have a section of the docs dedicated to installing the docker cli without the engine since people whose code won't work anymore will look for a solution and might not be aware that they need an apt-get or similar to replace the "download-cli". If you wish, that can be in another PR, but we should write that up before making a release.

LewisGaul:
Can we just link to https://docs.docker.com/engine/install/ and https://podman.io/docs/installation? Where in the docs were you thinking this should go?

gabrieldemarmiesse:
We should definitely link those.
I think the instructions should go to https://gabrieldemarmiesse.github.io/python-on-whales/docker_client/

I'm not sure linking is enough, we should additionally tell users that they don't need the engine and containerd and don't need to use systemd or systemctl to create the docker service.

The docker docs are not great if you just want to install the client so we should fill this gap. We just need to tell users what steps are not needed.

gabrieldemarmiesse (approved):
This looks really good, thanks! We'll pin an issue and display it loudly in the release notes, that should help with visibility. Thanks for everything!

LewisGaul:
Oh, there are some doc updates I was still working on sorry, I'll have to raise in a separate PR

gabrieldemarmiesse:
Something fancy could also be to re-enable CLI entrypoint (without typer) just to print the message that CLI isn't supported by python-on-whales anymore, so that people using it don't just get the "command not found" error. But that's slightly more work. What do you think?

LewisGaul:
Hm could do, unsure how many users would actually be relying on it...

Here's the follow-up PR: https://github.com/gabrieldemarmiesse/python-on-whales/pull/639

Issue #629, opened 2024-09-01
Missing support for passing multiple filters of the same type


LewisGaul:
Filtering can be performed when listing objects or pruning objects. In some components a `TypedDict` is used to indicate the supported filter types, while in other components a plain `Dict[str,str]` is used. In either case, this does not support something like `docker ps --filter label=foo=1 --filter label=bar`.

This was originally brought up in https://github.com/gabrieldemarmiesse/python-on-whales/pull/590, which specifically attempts to address the case of filtering by multiple labels, but the problem is wider than that (see discussion thread https://github.com/gabrieldemarmiesse/python-on-whales/pull/590#discussion_r1610269114).

This issue tracks fully resolving the root of the issue, which involves needing to change the APIs, as well as adding better typing to all components' filters.

LewisGaul:
My proposal for the new API is:

```python
docker.container.list(filters=[
    ("label", "foo"),
    ("label", "bar=1"),
    ("status", "created"),
    ("status", "running"),
    ("name", "prefix-*"),
])
```

This is implemented at https://github.com/gabrieldemarmiesse/python-on-whales/compare/master...LewisGaul:python-on-whales:filters-rework.

@gabrieldemarmiesse how should we handle the backwards incompatible change? Do we need to continue to accept `Dict[str,str]` on all these APIs and emit deprecation warnings?

gabrieldemarmiesse:
The new API looks good! Though I believe we shouldn't pull the rug on our users if we can avoid it, so as you say backward compatibility + depreciation warning seems the way to go here.

LewisGaul:
@xcvbn23 It's implemented in https://github.com/gabrieldemarmiesse/python-on-whales/pull/635, I've just added some tests so just needs to go through review.

PR #635, opened 2024-09-30
Rework filters to support passing multiple filters of the same type

LewisGaul:
Fixes https://github.com/gabrieldemarmiesse/python-on-whales/issues/629, supercedes https://github.com/gabrieldemarmiesse/python-on-whales/pull/590


gabrieldemarmiesse (approved):
That looks good! Many thanks! And sorry for the wait

PR #641, opened 2024-10-16
Add podman `is_infra` and `namespace` fields to `Container` object

LewisGaul:


LewisGaul:
Looks like we have two new unrelated CI failures...

Issue #594, opened 2024-05-25
Drop support for pydantic v1


LewisGaul:
The pydantic [version policy](https://docs.pydantic.dev/latest/version-policy/) says v1 support will be stopped at the end of June 2024, at which point we can also drop support.

- Remove any code that checks the pydantic version (search for `pydantic.VERSION`)
- Restrict the version in `requirements.txt` (which is also used by the packaging via `pyproject.toml`)
- Stop parameterising over pydantic versions in the CI

- [x] Wait for pydantic to drop support for v1

PR #642, opened 2024-10-19
Remove leftover Pydantic v1 handling

LewisGaul:
Missed handling for Pydantic v1, which had support removed as per https://github.com/gabrieldemarmiesse/python-on-whales/issues/594


gabrieldemarmiesse (approved):
This looks good. We can merge when the podman issue in the CI is fixed :)

PR #643, opened 2024-10-19
Set `defer_build=True` pydantic config, improving import time significantly

LewisGaul:
Partially addresses https://github.com/gabrieldemarmiesse/python-on-whales/issues/552. This config option is suggested in https://github.com/pydantic/pydantic/discussions/6748.

Before:
```
$time python -c "import python_on_whales"

real    0m0.405s
user    0m0.401s
sys     0m0.023s
```

After:
```
$time python -c "import python_on_whales"

real    0m0.206s
user    0m0.206s
sys     0m0.011s
```


gabrieldemarmiesse (approved):
I looked into it and indeed it makes a lot of sense. Thanks! We can merge this when the CI is fixed.

PR #651, opened 2024-11-19
Make secrets optional for BuildxCLI

### LINKED ISSUE #609: Buildx CLI shouldn't require secrets
The BuildxCLI build command requires the secrets argument to be set, but `docker buildx build` does not actually require  a `--secret` flag to be set to successfully run. Making secrets optional would bring this behavior in line with the docker buildx build command

Felixoid:
This PR should fix #609


gabrieldemarmiesse (approved):
Many thanks for the fix! Can you add a unit test which wouldn't work without this fix?

Felixoid:
Done. Didn't see them before the CI launched

Felixoid:
You know, if I remove the patch of `BuildxCLI`, it will still work. I think your code `add_args_iterable_or_single` is already very good, so I'd just add the test, and it's good to go.

gabrieldemarmiesse:
Yes, maybe I misunderstood the bug report

Felixoid:
Went through old PRs. Do you think it still worth to be merged?

PR #656, opened 2024-12-03
Support bake remote definitions

rcwbr:
Docker buildx bake supports [remote Bake file definitions](https://docs.docker.com/build/bake/remote-definition/#combine-local-and-remote-bake-definitions), but this functionality seems to be missing from the [existing API](https://gabrieldemarmiesse.github.io/python-on-whales/sub-commands/buildx/#python_on_whales.components.buildx.cli_wrapper.BuildxCLI.bake). This change intends to add support for this.

I was lazy about rendering the docs, so also threw in a change to save the rendered files from the CI pipeline as an artifact -- not sure if this would have too much impact on GitHub storage, but figured I'd propose the change.

A couple notes on the implementation:
1. The arg name I selected is "`context`" -- "`definition`" or "`target`" _might_ align better to [upstream docs](https://docs.docker.com/reference/cli/docker/buildx/bake/), though neither is clearly specified in docs and all are overloaded 😬 Very happy to change to a better name if there is one!
1. The `context` arg is directly `append`ed to the `full_cmd`, meaning the content of the arg could technically be _any_ valid bake CLI arg. Validation could prevent this, but didn't seem too important.
1. In the test, I locked the remote ref to the `v0.74.0` tag. This seems slightly awkward, but should at least be safe since its immutable.

Please let me know if I'm missing anything, or if anything needs to be changed!

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -22,6 +22,10 @@ jobs:
           cp docs/mkdocs.yml ./
       - name: Render
         run: uv run mkdocs build
+      - uses: actions/upload-artifact@v4
+        with:
+          name: Docs
+          path: site
 
   build-linux-lint:
     runs-on: ubuntu-24.04
--- a/python_on_whales/components/buildx/cli_wrapper.py
+++ b/python_on_whales/components/buildx/cli_wrapper.py
@@ -126,6 +126,7 @@ class BuildxCLI(DockerCLICaller):
         set: Dict[str, str] = {},
         variables: Dict[str, str] = {},
         stream_logs: bool = False,
+        remote_definition: Union[str, None] = None,
     ) -> Union[Dict[str, Dict[str, Dict[str, Any]]], Iterator[str]]:
         """Bake is similar to make, it allows you to build things declared in a file.
 
@@ -148,6 +149,7 @@ class BuildxCLI(DockerCLICaller):
             set: A list of overrides in the form `"targetpattern.key=value"`.
             variables: A dict containing the values of the variables defined in the
                 hcl file. See <https://github.com/docker/buildx#hcl-variables-and-functions>
+            remote_definition: Remote context in which to find bake files
 
         # Returns
             The configuration used for the bake (files merged + override with
@@ -197,6 +199,8 @@ class BuildxCLI(DockerCLICaller):
         for file in to_list(files):
             full_cmd.add_simple_arg("--file", file)
         full_cmd.add_args_iterable_or_single("--set", format_mapping_for_cli(set))
+        if remote_definition is not None:
+            full_cmd.append(remote_definition)
         targets = to_list(targets)
         env = dict(variables)
         if print:
--- a/tests/python_on_whales/components/buildx/test_buildx_cli_wrapper.py
+++ b/tests/python_on_whales/components/buildx/test_buildx_cli_wrapper.py
@@ -656,6 +656,55 @@ def test_bake_with_variables_2(only_print, monkeypatch):
 
 @pytest.mark.usefixtures("with_docker_driver")
 @pytest.mark.usefixtures("change_cwd")
+@pytest.mark.parametrize("only_print", [True, False])
+def test_bake_with_remote_definition(only_print):
+    config = docker.buildx.bake(
+        print=only_print,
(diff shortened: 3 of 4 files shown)


gabrieldemarmiesse:
Many thanks for the PR! I'll take a look when I have the time :)

tests/python_on_whales/components/buildx/test_buildx_cli_wrapper.py
```diff
@@ -654,6 +654,33 @@ def test_bake_with_variables_2(only_print, monkeypatch):
...
 
 
+@pytest.mark.usefixtures("with_docker_driver")
+@pytest.mark.usefixtures("change_cwd")
+@pytest.mark.parametrize("only_print", [True, False])
+def test_bake_with_remote_context(only_print):
+    config = docker.buildx.bake(
+        print=only_print,
+        context="https://github.com/gabrieldemarmiesse/python-on-whales.git#v0.74.0:tests/python_on_whales/components/bake_tests",
+    )
```
gabrieldemarmiesse, line 664: Can you also add a test where you specify a target to make sure the arguments are passed in the right order?

python_on_whales/components/buildx/cli_wrapper.py
```diff
@@ -126,6 +126,7 @@ def bake(
         set: Dict[str, str] = {},
         variables: Dict[str, str] = {},
         stream_logs: bool = False,
+        context: Union[str, None] = None,
```
gabrieldemarmiesse, line 129: The issue I have with context is that `context` already exists: https://docs.docker.com/build/bake/contexts/ and it's slight different.

What about being explicit? Unless I'm missing something, we can just use "remote_definition", it's the term used in the docker documentation.

.github/workflows/python-package.yml
```diff
@@ -22,6 +22,10 @@ jobs:
           cp docs/mkdocs.yml ./
       - name: Render
         run: uv run mkdocs build
+      - uses: actions/upload-artifact@v4
```
gabrieldemarmiesse, line 25: Not mandatory, but since this is an awesome feature, you can add it to the `CONTRIBUTING.md` :) so contributors don't have to build the docs themselves. Just say 
- where to click to get the files
- unzip it
- open index.html with the browser

Again it's optional since you contributed this feature and it's a net improvement, but if you want to go the extra mile, I'd appreciate a note in the contributing guide.

gabrieldemarmiesse (changes requested):
Very very nice PR, easy to read, straight to the point. ❤️  A few changes and we can merge this.

tests/python_on_whales/components/buildx/test_buildx_cli_wrapper.py
rcwbr: Good call, [added](https://github.com/gabrieldemarmiesse/python-on-whales/pull/656/files#diff-e30722b601498d80a6278c8eb440452e25f646641ded295374cd13402ff58fd9R687)

python_on_whales/components/buildx/cli_wrapper.py
rcwbr: Yeah `remote_definition` seems good, changed to that.

The interesting behavior with `context` is that by default, the contents of the `remote_context` become the value of Bake target `context`s (not `contexts`), [for example in the test definition](https://github.com/gabrieldemarmiesse/python-on-whales/pull/656/files#diff-e30722b601498d80a6278c8eb440452e25f646641ded295374cd13402ff58fd9R669). However, this is only the _default_; if the Bake file defines a `context`, it may be unrelated to this value.

TBH I think the whole idea of this Bake functionality needs some clarification/cleanup from Docker folks, but until that happens I think this is the best we can do!

.github/workflows/python-package.yml
rcwbr: No problem, [added](https://github.com/gabrieldemarmiesse/python-on-whales/pull/656/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R43)!

rcwbr:
> Very very nice PR, easy to read, straight to the point. ❤️ A few changes and we can merge this.

Thank you, and thanks for the awesome library!!! Should be ready for re-review now if you'd like

gabrieldemarmiesse (approved):
Very nice!

gabrieldemarmiesse:
Hi @rcwbr sorry I didn't push a release earlier, it's quite fast on my side of things, I just forgot. I'll do it tonight :)

Issue #659, opened 2024-12-23
Support for --entrypoint and --env in docker.compose.run()


www84:
This library is great, thank you for creating it! I found out that --entrypoint and --env is currently not supported for docker.compose.run(), which I desperately need. I just finished adding these options as well as tests on my fork but then I saw that each commit corresponds to some issue, so I'm opening one. If I'm not following the intended process, please let me know.

gabrieldemarmiesse:
We don't have a formal process. Feel free to open a pull request. If possible, prefer to open one pull request per feature added as it's easier to merge and rollback if needed :) thank you for your interest in making this project better!

www84:
Noted! The current PR is in test, so I'm waiting until the tests are finished. After that I will close the MR and split it into two, one for each option.

www84:
I create a separate issue for --env.

PR #661, opened 2024-12-23
Add entrypoint option to docker.compose.run() (#659)

www84:
PR for the entrypoint part of #659, a second PR follows if this is merged successfully.

### DIFF
--- a/python_on_whales/components/compose/cli_wrapper.py
+++ b/python_on_whales/components/compose/cli_wrapper.py
@@ -663,7 +663,7 @@ class ComposeCLI(DockerCLICaller):
         command: List[str] = [],
         build: bool = False,
         detach: bool = False,
-        entrypoint: Optional[List[str]] = None,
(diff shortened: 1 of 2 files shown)


python_on_whales/components/compose/cli_wrapper.py
```diff
@@ -663,7 +663,7 @@ def run(
         command: List[str] = [],
         build: bool = False,
         detach: bool = False,
-        # entrypoint: Optional[List[str]] = None,
+        entrypoint: Optional[List[str]] = None,
```
gabrieldemarmiesse, line 666:
```suggestion
        entrypoint: Optional[str] = None,
```
In the tests you're passing a string, so I believe that should be wrong to expect a `List[str]`

gabrieldemarmiesse (changes requested):
Nice pull request, just a small fix, but I should be able to push it myself.

gabrieldemarmiesse (approved):

Issue #662, opened 2024-12-25
Support for --volume in docker.compose.run()


www84:
Support for the option --volume in `docker compose run`.

PR #663, opened 2024-12-25
Add volumes option to docker.compose.run() (#662)

www84:
Implemented option for --volume in docker.compose.run() including tests.

### DIFF
--- a/python_on_whales/components/compose/cli_wrapper.py
+++ b/python_on_whales/components/compose/cli_wrapper.py
@@ -8,6 +8,7 @@ from typing import Any, Dict, Iterable, List, Optional, Tuple, Union, overload
 from typing_extensions import Literal
 
 import python_on_whales.components.container.cli_wrapper
+import python_on_whales.components.volume.cli_wrapper
 from python_on_whales.client_config import DockerCLICaller
 from python_on_whales.components.compose.models import ComposeConfig, ComposeProject
 from python_on_whales.utils import (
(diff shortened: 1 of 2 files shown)


www84:
Okay, my IDE told me on commit that test_compose.py has Windows line breaks, but I thought that I accidentally switched this one file to Windows line breaks it in my IDE, so I changed it to Linux line breaks. Let me quickly fix that back to Windows line breaks.

www84:
It's back to Windows line breaks. The question is, thought, if this is desired or if this is an oversight and all files should have Linux line breaks.

gabrieldemarmiesse (approved):
Very nice pull request! Thanks!

gabrieldemarmiesse:
> It's back to Windows line breaks. The question is, thought, if this is desired or if this is an oversight and all files should have Linux line breaks.

Linux line breaks are preferred but I don't want to have to check it at every pull request so I don't ask people to change it if I notice it. A github action checking that all line breaks are linux style would be the best, as it would give quick feedback to contributors but it's not a big deal if we don't have it.

www84:
> Linux line breaks are preferred but I don't want to have to check it at every pull request so I don't ask people to change it if I notice it. A github action checking that all line breaks are linux style would be the best, as it would give quick feedback to contributors but it's not a big deal if we don't have it.
Alright. I agree that it's not a big issue. One last PR coming up in a few minutes. Thank you for reviewing at this time of the year. :+1:

PR #668, opened 2025-01-23
Multi-Node-Builder support

rico132:
buildx/cli_wrapper: add the --append flag to the create command
buildx/cli_wrapper: update platforms() and status() properties, add nodes() property
models.py: add class BuilderNode
models.py: rewrite from_str method so it also parses BuilderNodes

### DIFF
--- a/python_on_whales/components/buildx/models.py
+++ b/python_on_whales/components/buildx/models.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+
 from dataclasses import dataclass, field
 from typing import List
 
--- a/python_on_whales/components/buildx/models.py
+++ b/python_on_whales/components/buildx/models.py
@@ -3,7 +3,6 @@ from dataclasses import dataclass, field
 from typing import List
 
 
-
 @dataclass
 class BuilderInspectResult:
     name: str
@@ -41,6 +40,7 @@ class BuilderInspectResult:
                     result_dict["nodes"].append(node)
                     node = BuilderNode()
         return cls(**result_dict)
+
 
 class BuilderNode:
     name: str = ""
--- a/python_on_whales/components/buildx/models.py
+++ b/python_on_whales/components/buildx/models.py
@@ -4,27 +4,49 @@ from dataclasses import dataclass, field
 from typing import List
 
 
+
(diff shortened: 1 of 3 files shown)


gabrieldemarmiesse:
I fixed the formatting with `uv run --with pre-commit pre-commit run --all-files`

python_on_whales/components/buildx/models.py
```diff
@@ -8,23 +8,45 @@
...
+                if line.startswith("Name:"):
+                    node.name = line.split(":")[1].strip()
+                if line.startswith("Status:"):
+                    node.status = line.split(":")[1].strip()
+                if line.startswith("Platforms:"):
+                    platforms = line.split(":")[1].strip()
+                    if platforms:
+                        node.platforms = platforms.split(", ")
+                    result_dict["nodes"].append(node)
+                    node = BuilderNode()
```
gabrieldemarmiesse, line 42: Rather than doing all this parsing, I believe we should leverage `docker buildx ls --format "{{json . }}"`, I will look into it.

rico132: Yes, but you mentioned that ```docker buildx ls --format "{{json . }}"``` loses the information of what builder is being used, so I parsed it from ```docker buildx inspect```

gabrieldemarmiesse: Maybe we can grab the builder name from `docker buildx inspect`, and then run `docker buildx ls --format "{{json . }}"` , we loop over the results and take the entry corresponding to our builder. Then parsing the json is easy peasy and can be loaded into a pydantic model

rico132: Do you want to do this in the [_fetch_and_parse_inspect_result](https://github.com/gabrieldemarmiesse/python-on-whales/blob/b6b72a6efe9f3029bff6f2f0f8a0a6b080623581/python_on_whales/components/buildx/cli_wrapper.py#L60) function? Or do you want to create another function to do this?

gabrieldemarmiesse:
Closing in favor of #674 which has better integration with the latest versions of buildx. Sorry for the delay @rico132

PR #676, opened 2025-05-16
Add events method to stream Docker Compose events

pietroagazzi:
Implemented a new `events` method in the `ComposeCLI` class to stream real-time Docker Compose events for specified services or all services in a project. Added corresponding `ComposeEvent` model and a test case to verify the functionality.

### DIFF
--- a/tests/python_on_whales/components/test_compose.py
+++ b/tests/python_on_whales/components/test_compose.py
@@ -12,7 +12,7 @@ import pytz
 
 import python_on_whales
 from python_on_whales import DockerClient, DockerException
-from python_on_whales.components.compose.models import ComposeConfig
+from python_on_whales.components.compose.models import ComposeConfig, ComposeEvent
 from python_on_whales.exceptions import NoSuchImage
 from python_on_whales.test_utils import get_all_jsons
 from python_on_whales.utils import PROJECT_ROOT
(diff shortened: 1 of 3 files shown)


tests/python_on_whales/components/test_compose.py
```diff
@@ -894,6 +894,37 @@ def test_compose_logs_follow():
...
+                break
+    except KeyboardInterrupt:
+        pass
+
+    assert len(events) > 0
+    assert all(
+        isinstance(event, ComposeEvent) and event.service == "unhealthy_service"
+        for event in events
+    )
+
```
gabrieldemarmiesse, line 924: Could you check a bit more info? Like the other attributes for example, and the number of event collected (we make sure the for loop actually ran). Notably, it would make sure the other attributes are not `None`, and that we didn't make any typos in the pydantic model for the events.

tests/python_on_whales/components/test_compose.py
```diff
@@ -894,6 +894,37 @@ def test_compose_logs_follow():
...
+    events_iterator = docker.compose.events(["unhealthy_service"])
+
+    events = []
+    try:
+        for event in events_iterator:
+            events.append(event)
+            if len(events) >= 3:  # Collect a few events for testing
+                break
+    except KeyboardInterrupt:
```
gabrieldemarmiesse, line 916: I'm not sure why the try except is there. If there is a keyboard interrupt, why would we want to swallow the exception?

gabrieldemarmiesse (changes requested):
Awesome stuff! I wasn't aware that compose events existed, and I can see how useful it can be for monitoring purposes!

tests/python_on_whales/components/test_compose.py
pietroagazzi: You're right, the try/except block is not necessary here. I'll remove it.

gabrieldemarmiesse (approved):
Great PR!

gabrieldemarmiesse:
@pietroagazzi We'll give ourselves some time, and if nobody reported any bug, we'll do a release on Pypi around friday

PR #680, opened 2025-06-18
Allow customization of `--isolation` and `--quiet` in `legacy_build`

makuchma:
The legacy builder is still useful for Windows scenarios not yet fully supported by BuildKit.

This change opens up a couple options for the users of `legacy_build`:

- Add support to `--isolation`: this is necessary to enable Hyper-V isolation.
- Allow `--quiet` to be unset: for easier debugging.

### DIFF
--- a/python_on_whales/components/image/cli_wrapper.py
+++ b/python_on_whales/components/image/cli_wrapper.py
@@ -252,6 +252,7 @@ class ImageCLI(DockerCLICaller):
         tags: Union[str, Iterable[str]] = (),
         target: Optional[str] = None,
         isolation: Optional[str] = None,
+        quiet: bool = True,
     ) -> python_on_whales.components.image.cli_wrapper.Image:
         """Build a Docker image with the old Docker builder (meaning not using buildx/buildkit)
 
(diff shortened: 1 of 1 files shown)


python_on_whales/components/image/cli_wrapper.py
```diff
@@ -299,13 +303,25 @@ def legacy_build(
...
+        else:
+            for _, line in stream_stdout_and_stderr(full_cmd):
+                try:
+                    line = line.decode().rstrip()
+                except UnicodeDecodeError:
+                    line = str(line)
+                print(line)
+                if "Successfully built" in line:
+                    image_id = line.split(" ")[-1]
+            return docker_image.inspect(image_id)
```
gabrieldemarmiesse, line 324: Is this else branch necessary?

python_on_whales/components/image/cli_wrapper.py
```diff
@@ -251,6 +251,8 @@ def legacy_build(
         pull: bool = False,
         tags: Union[str, Iterable[str]] = (),
         target: Optional[str] = None,
+        isolation: Optional[str] = None,
+        quiet: bool = True,
```
gabrieldemarmiesse, line 255: Why is it quiet by default?

gabrieldemarmiesse (commented):
Thanks for the pull request! I have a few questions related to it.

python_on_whales/components/image/cli_wrapper.py
makuchma: I think so. We want to print the build output and capture the image ID. The line with the image ID will be slightly different than in the `if` version (e.g: `sha256:94eec2b81d031b31f1230ca017274a9427cdf18b5824bf3714b9ba697e5f5236` vs `Successfully built 94eec2b81d03`). Do you see a better way of doing this?

python_on_whales/components/image/cli_wrapper.py
makuchma: The current version has `--quiet` enabled so I thought it would be better to keep it like that by default to keep it compatible and avoid problems for existing clients. :-)

PR #691, opened 2025-10-23
Add 'pull' argument to compose.create

g-arjones:
The `--pull` option is available on both `docker compose up` and `docker compose create` but for some reason it was not implemented in the "compose" CLI wrapper.

### DIFF
--- a/python_on_whales/components/compose/cli_wrapper.py
+++ b/python_on_whales/components/compose/cli_wrapper.py
@@ -195,7 +195,7 @@ class ComposeCLI(DockerCLICaller):
             no_build: Don't build an image, even if it's missing.
             no_recreate: If containers already exist, don't recreate them.
                 Incompatible with `force_recreate=True`.
-            pull: Pull image before running (“always”|”missing”|”never”|”build”).
+            pull: Pull image before running ("always"|"missing"|"never"|"build").
             stream_logs: If `False` this function returns None. If `True`, this
                 function returns an Iterable of `Tuple[str, bytes]` where the first element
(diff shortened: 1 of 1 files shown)


python_on_whales/components/compose/cli_wrapper.py
```diff
@@ -193,6 +195,7 @@ def create(
             no_build: Don't build an image, even if it's missing.
             no_recreate: If containers already exist, don't recreate them.
                 Incompatible with `force_recreate=True`.
+            pull: Pull image before running (“always”|”missing”|”never”|”build”).
```
gabrieldemarmiesse:
```suggestion
            pull: Pull image before running ("always"|"missing"|"never"|"build").
```
Let's use normal quotes

gabrieldemarmiesse (changes requested):

python_on_whales/components/compose/cli_wrapper.py
g-arjones: Sounds good to me, although I copied it from the `up()` docs for consistency...

python_on_whales/components/compose/cli_wrapper.py
```diff
@@ -205,6 +208,7 @@ def create(
         full_cmd.add_flag("--force-recreate", force_recreate)
         full_cmd.add_flag("--no-build", no_build)
         full_cmd.add_flag("--no-recreate", no_recreate)
+        full_cmd.add_simple_arg("--pull", pull)
```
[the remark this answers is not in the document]
g-arjones: Amended to patch the test case.

PR #709, opened 2026-03-19
Add support for --disable-dns when creating a network (podman)

Hasnainz:
`podman network create` has an option to disable creating a dns process when making bridge networks, see documentation [here](https://docs.podman.io/en/latest/markdown/podman-network-create.1.html). As far as I can tell docker doesn't have this flag?

### DIFF
--- a/python_on_whales/components/network/cli_wrapper.py
+++ b/python_on_whales/components/network/cli_wrapper.py
@@ -202,6 +202,7 @@ class NetworkCLI(DockerCLICaller):
         self,
         name: str,
         attachable: bool = False,
+        disable_dns: bool = False,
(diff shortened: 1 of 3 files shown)


python_on_whales/components/network/cli_wrapper.py
LewisGaul, line 219: I realise there's a gap in that the docstring doesn't cover most of the options, but could you add this one since it's important we note that this is only valid for Podman.

LewisGaul, line 1: Can we add a simple testcase, even if it just creates the network with the flag and checks `network.dns_enabled` (does this need adding to the model?)

PR #710, opened 2026-03-25
Add stdin support for interactive containers

BrandonHaynes:
# Add stdin support for interactive streaming containers

## Summary

Allow `interactive=True` and `stream=True` to be used together in `docker.execute()`. This enables programmatic bidirectional communication with a running container — writing to stdin while streaming stdout/stderr.

Previously, combining these flags raised a `ValueError` because the original implementation treated interactive mode as a TTY-only feature. However, there are valid use cases for piping stdin to a container process while streaming its output — for example, feeding input to a long-running subprocess and reacting to its output line-by-line.

### `python_on_whales/utils.py`

- **`ProcessStream`** — New minimal wrapper class around the streaming iterator that also exposes the subprocess's `stdin` handle. Iterating over it yields `(source, line)` tuples just like before. When `pipe_stdin=False`, `stdin` is `None`.
- **`_start_process()`** — Extracted from `stream_stdout_and_stderr`. Starts the subprocess and wires up reader threads. Accepts a `pipe_stdin` flag to optionally open `stdin=PIPE`.
- **`_iter_process()`** — Extracted from `stream_stdout_and_stderr`. Generator that yields `(source, line)` tuples from the reader queue, then checks exit status.
- **`stream_stdout_and_stderr()`** — Now accepts `pipe_stdin` and always returns a `ProcessStream`.

### `python_on_whales/components/container/cli_wrapper.py`

- Removed the `ValueError` for `interactive=True, stream=True`.
- Passes `pipe_stdin=interactive` to `stream_stdout_and_stderr()`.
- Updated docstring to describe the new behavior.

### Tests

- **`test_utils.py`** — 3 new unit tests:
  - Interactive single-line write/read via `ProcessStream`
  - Interactive multi-line write/read
  - Non-interactive mode returns `ProcessStream` with `stdin=None`
- **`test_container.py`** — 1 new integration test:
  - `test_execute_interactive_stream` — runs `cat` in a container, writes to `proc.stdin`, and verifies the echoed output

## Usage

```python
container = docker.run("ubuntu", ["sleep", "infinity"], detach=True)

proc = docker.execute(container, ["cat"], interactive=True, stream=True)
proc.stdin.write(b"hello\n")
proc.stdin.close()

for source, line in proc:
    print(source, line)
# stdout b'hello\n'
```

## Backward Compatibility

No breaking changes. `stream_stdout_and_stderr()` now returns a `ProcessStream` instead of a bare generator, but `ProcessStream` implements `__iter__` so all existing iteration patterns continue to work. The `pipe_stdin` parameter defaults to `False`.

### DIFF
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -1002,8 +1002,9 @@ class ContainerCLI(DockerCLICaller):
             env_files: Read one or more files of environment variables
             interactive: Leave stdin open during the duration of the process
                 to allow communication with the parent process.
-                Currently only works with `tty=True` for interactive use
-                on the terminal.
+                Can be combined with `stream=True` to get an
+                `ProcessStream` that allows writing to stdin while
+                iterating over stdout/stderr.
             preserve_fds: The number of additional file descriptors to pass
                 through to the container. Only supported by podman.
             privileged: Give extended privileges to the container.
@@ -1041,12 +1042,6 @@ class ContainerCLI(DockerCLICaller):
         full_cmd.add_args_mapping("--env", envs)
         full_cmd.add_args_iterable_or_single("--env-file", env_files)
 
-        if interactive and stream:
-            raise ValueError(
-                "You can't set interactive=True and stream=True at the same"
-                "time. Their purpose are not compatible."
-            )
-
         if tty and stream:
             raise ValueError(
                 "You can't set tty=True and stream=True at the same"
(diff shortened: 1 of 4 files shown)


BrandonHaynes:
@gabrieldemarmiesse any signal on whether you'd consider taking this into `main`?  We have an internal use case blocked on this change.  

cc @kvemani

gabrieldemarmiesse:
Apologies for the delay, this seems interesting, I'll take a look :)  I've been quite busy recently but I'll find a way

gabrieldemarmiesse:
Thanks for the pull request!

PR #627, opened 2024-09-01
Update container component to accept args as Iterable and Mapping

LewisGaul:
Follows https://github.com/gabrieldemarmiesse/python-on-whales/pull/595 and https://github.com/gabrieldemarmiesse/python-on-whales/pull/599, working on https://github.com/gabrieldemarmiesse/python-on-whales/issues/584.

PR #714, opened 2026-05-07
Update secret component to accept args as Iterable and Mapping

nanookclaw:
Follows #595, #599, #627, working on #584.

Applies the same `Iterable`/`Mapping` treatment to the secret component:

- `SecretCLI.create(labels=...)`: `Dict[str, str]` -> `Mapping[str, str]`
- `SecretCLI.inspect(x)`: `List[str]` -> `Iterable[str]`, with `@overload` so a `str` argument is typed back to `Secret` and an iterable to `List[Secret]` (matches the pod component shape from #599).
- `SecretCLI.remove(x)`: `List[ValidSecret]` -> `Iterable[ValidSecret]`. The empty-input shortcut moves from `x == []` to `to_list(x)` + emptiness check, since iterables are not equality-comparable to a list and would otherwise have raised on a generator.

`inspect`'s runtime branch flips from `isinstance(x, list)` to `isinstance(x, str)`, which correctly accepts any non-string iterable (tuple, set, generator, mapping keys) instead of silently treating them as a single id. `to_list` already handles `str`/`bytes`/`Iterable` correctly for `remove`.

No public-API breakage: callers passing lists keep working; type-checkers stop complaining about invariant `List[ValidSecret]` arguments (the original mypy issue in #584).

### DIFF
--- a/python_on_whales/components/secret/cli_wrapper.py
+++ b/python_on_whales/components/secret/cli_wrapper.py
@@ -110,9 +110,7 @@ class SecretCLI(DockerCLICaller):
     @overload
     def inspect(self, x: Iterable[str]) -> List[Secret]: ...
 
-    def inspect(
-        self, x: Union[str, Iterable[str]]
-    ) -> Union[Secret, List[Secret]]:
+    def inspect(self, x: Union[str, Iterable[str]]) -> Union[Secret, List[Secret]]:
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse (approved):
Thanks! that makes sense, it's a nice improvement :)

PR #715, opened 2026-05-08
feat(container): add encoding and errors params to logs() (#608)

### LINKED ISSUE #608: Should support passing encoding when getting container output
See https://github.com/gabrieldemarmiesse/python-on-whales/pull/607#issue-2401828786

> It might be good to support the user passing encoding and errors through (much like [subprocess.run()](https://docs.python.org/3/library/subprocess.html#subprocess.run))

nanookclaw:
Closes #608.

Adds `encoding` and `errors` keyword-only parameters to `ContainerCLI.logs()` and the `Container.logs()` proxy. Both default to current behavior (`utf-8` / `"strict"`), so the change is backwards-compatible.

The parameters mirror `subprocess.run()`'s naming, as suggested on #607. They only affect the `str` return path; when `stream=True` the iterator continues to yield raw `bytes` and the caller decodes per chunk.

## Test

Added one parametrized (docker/podman) test that runs `printf "\\xff\\xfeabc"` in busybox and asserts:
- `logs(c, errors="replace")` returns a string containing the U+FFFD replacement character and ending in `abc`
- `logs(c)` (default `errors="strict"`) raises `UnicodeDecodeError`

## Scope

`compose.logs`, `service.logs`, and `pod.logs` were intentionally left untouched and can follow if you want the pattern propagated.

### DIFF
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -358,6 +358,8 @@ class Container(ReloadableObjectFromJson):
         tail: Optional[int] = None,
         timestamps: bool = False,
         until: Union[None, datetime, timedelta] = None,
+        encoding: Optional[str] = None,
+        errors: str = "strict",
     ) -> str:
         """Returns the logs of the container
 
@@ -371,6 +373,8 @@ class Container(ReloadableObjectFromJson):
             tail=tail,
             timestamps=timestamps,
             until=until,
+            encoding=encoding,
+            errors=errors,
         )
 
     def pause(self) -> None:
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse (changes requested):
Thank you for the PR! I don't think this feature is worth the additional complexity of the api. Users can indeed use the stream option and handle the bytes however they want, which is the recommended path if you have a specific encoding you're using.

LewisGaul:
I would argue that anywhere arbitrary bytes are decoded there should be support for specifying the decoding to use - who are we to say that container output should be utf-8? This does seem to be the case in all of `open()`, `pathlib.Path.read_text()`, `subprocess.run()`... in the stdlib.

PR #716, opened 2026-06-07
Support cpuset strings on container create/run/update and pod create (#626)

### LINKED ISSUE #626: Inconsistent representation of cpusets
`ContainerCLI.create()` and `ContainerCLI.run()` support `cpuset_cpus` and `cpuset_mems` as `List[int]`, which is converted to a CLI arg using `",".join(...)`.

`ContainerCLI.update()` supports `cpuset_cpus` and `cpuset_mems` as `List[int]` in the type signature but then converts to a CLI arg simply by stringifying (implicitly in `utils.run()`), which I assume will always fail.

In reality the CLI accepts a cpuset representation, which includes: a range, e.g. `"0-3"`, a selection, e.g. `"0,2"`, a combination of both, e.g. `"0-3,7"`.

I'd suggest it'd make most sense just to support the user specifying this as a string, rather than only allowing the "selection" format via a list. This would mean fixing the type annotation on `update()` and a backwards incompatible change on `create()` and `run()`.

AliBassam:
Closes #626.

## Summary

`cpuset_cpus` / `cpuset_mems` on `container.create()`, `container.run()` and `pod.create()` only accepted `List[int]`, so users couldn't pass the range/mixed cpuset strings the Docker CLI actually supports (`"0-3"`, `"0,2"`, `"0-3,7"`). This PR widens the accepted types and adds a small `format_cpuset` helper that passes strings through unchanged and joins integer iterables.

`container.update()` was already typed `Optional[str]` on master but its docstring still claimed it required a list — that's corrected too.

## Design note

Issue #626 suggested making this **string-only** (a backwards-incompatible change). I went with `Union[List[int], str, None]` instead to keep existing list-based callers working. Happy to drop `List[int]` and make it string-only with a deprecation if you'd prefer — let me know and I'll push a follow-up commit.

## Changes

- New `format_cpuset()` util in `python_on_whales/utils.py`
- `container.create()`, `container.run()`: widened `cpuset_cpus` / `cpuset_mems` to `Union[List[int], str, None]`, switched call-sites to `format_cpuset`
- `pod.create()`: widened `cpuset_cpus` to `Union[Iterable[int], str, None]`
- `container.update()`: docstring corrected (signature was already `Optional[str]`)
- Unit tests for `format_cpuset` covering None / list / empty / string / generator

### DIFF
--- a/python_on_whales/components/container/cli_wrapper.py
+++ b/python_on_whales/components/container/cli_wrapper.py
@@ -48,6 +48,7 @@ from python_on_whales.utils import (
     ValidPath,
     ValidPortMapping,
     custom_parse_object_as,
+    format_cpuset,
     format_port_arg,
     format_signal_arg,
     format_time_arg,
@@ -459,8 +460,8 @@ class RunArgs(TypedDict, total=False):
     cpu_rt_runtime: Optional[int]
     cpu_shares: Optional[int]
     cpus: Optional[float]
-    cpuset_cpus: Optional[List[int]]
-    cpuset_mems: Optional[List[int]]
+    cpuset_cpus: Union[List[int], str, None]
+    cpuset_mems: Union[List[int], str, None]
     devices: Iterable[str]
     device_cgroup_rules: Iterable[str]
     device_read_bps: Iterable[str]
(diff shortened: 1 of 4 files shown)


gabrieldemarmiesse (approved):
Looks good to me, thanks!

PR #717, opened 2026-06-22
Support for 'docker compose publish' command

MisterOwlPT:
Hi!

This PR adds support for the Docker Compose `publish` command, allowing users to version and distribute Compose YAML files using automation scripts and `python-on-whales`. 

I've recently discovered this Docker Compose feature and realized that it is quite useful for some workflows at work — and since we're already using `python-on-whales` extensively, adding support to it made perfect sense. Besides all, it also increases the compliance levels of the library with the Docker Compose CLI.

Here's the `help` for the mentioned command:

```bash
$ docker compose publish --help
Usage:  docker compose publish [OPTIONS] REPOSITORY[:TAG]

Publish compose application
```

As an usage example, consider the following Compose file:

```yaml
# example_deployment.yml
---
services:
  hello:
    image: alpine:latest
    environment:
      NAME: ${NAME}
    command: sh -c 'while true; do echo "$$NAME"; sleep 5; done'
```

With the proposed changes, users can now distribute their Compose applications as follows:

```python
>>> from python_on_whales import DockerClient
>>> # ...
>>> docker = DockerClient(compose_files=["example_deployment.yml"])
>>> docker.compose.publish("docker.io/username/my-application:v0.0.1", yes=True)
```

All available command options are supported (e.g., the flag `--yes` in the example above to bypass interactive prompts in automated contexts).

Also with the proposed changes, users may download and use published applications as follows:

```python
>>> import os
>>> from python_on_whales import DockerClient
>>> docker = DockerClient(compose_files=["oci://docker.io/username/my-application:v0.0.1"])
>>> # ...
>>> os.environ["NAME"] = "DODO"
>>> # ...
>>> docker.compose.up(yes=True)
[+] Running 1/1
 ✔ Container components-hello-1  
 Created                   0.0s 
Attaching to hello-1
hello-1  | DODO
hello-1  | DODO
hello-1  | DODO
```

Additionally, I've added the missing flag `--yes` to existing command `docker.compose.up`.

The PR includes unit tests for the newly added logic and documentation as per the Docker Compose CLI.

Let me know what you think of it all.
Thanks!

### DIFF
--- a/python_on_whales/components/compose/cli_wrapper.py
+++ b/python_on_whales/components/compose/cli_wrapper.py
@@ -531,6 +531,32 @@ class ComposeCLI(DockerCLICaller):
             for proj in json.loads(run(full_cmd))
         ]
 
+    def publish(
+        self,
+        repository: str,
+        oci_version: Optional[str] = None,
+        resolve_image_digests: bool = False,
+        with_env: bool = False,
+        yes: bool = False,
+    ) -> None:
+        """Publish compose application.
+
+        Parameters:
+            repository: Repository to publish compose application
+            oci_version: OCI image/artifact specification version (automatically determined by default)
+            resolve_image_digests: Pin image tags to digests
+            with_env: Include environment variables in the published OCI artifact
+            yes: Assume "yes" as answer to all prompts
+        """
+        full_cmd = self.docker_compose_cmd + ["publish"]
+        if oci_version is not None:
+            full_cmd.add_simple_arg("--oci-version", oci_version)
+        full_cmd.add_flag("--resolve-image-digests", resolve_image_digests)
+        full_cmd.add_flag("--with-env", with_env)
+        full_cmd.add_flag("--yes", yes)
+        full_cmd.append(repository)
+        run(full_cmd, capture_stderr=False, capture_stdout=False)
+
     @overload
     def pull(
         self,
@@ -969,6 +995,7 @@ class ComposeCLI(DockerCLICaller):
         stream_logs: Literal[True] = ...,
         wait_timeout: Optional[int] = ...,
         dependencies: bool = True,
+        yes: bool = ...,
     ) -> Iterable[Tuple[str, bytes]]: ...
 
     @overload
@@ -995,6 +1022,7 @@ class ComposeCLI(DockerCLICaller):
         stream_logs: Literal[False] = ...,
         wait_timeout: Optional[int] = ...,
         dependencies: bool = True,
+        yes: bool = ...,
     ) -> None: ...
 
     def up(
@@ -1020,6 +1048,7 @@ class ComposeCLI(DockerCLICaller):
         stream_logs: bool = False,
         wait_timeout: Optional[int] = None,
         dependencies: bool = True,
+        yes: bool = False,
     ):
         """Start the containers.
 
@@ -1066,6 +1095,7 @@ class ComposeCLI(DockerCLICaller):
                 not familiar with the streaming of logs in Python-on-whales.
             wait_timeout: Maximum duration to wait for the project to be running|healthy
             dependencies: Also start linked services.
+            yes: If `True` assume "yes" as answer to all prompts and run non-interactively.
         """
         if quiet and stream_logs:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


python_on_whales/components/compose/cli_wrapper.py
```diff
@@ -531,6 +531,32 @@ def ls(
...
+            resolve_image_digests: Pin image tags to digests
+            with_env: Include environment variables in the published OCI artifact
+            yes: Assume "yes" as answer to all prompts
+        """
+        full_cmd = self.docker_compose_cmd + ["publish"]
+        if oci_version is not None:
+            full_cmd.add_simple_arg("--oci-version", oci_version)
+        full_cmd.add_flag("--resolve-image-digests", resolve_image_digests)
+        full_cmd.add_flag("--with-env", with_env)
+        full_cmd.add_flag("--yes", yes)
```
gabrieldemarmiesse: Shouldn't yes be always true? I think we don't want the program to ever stop to ask something to the user, that seems like bad UX. In which situation will the docker CLI stop to ask the user permission?

gabrieldemarmiesse (commented):
Many thanks for the PR! it looks good to me. Just a question related to the --yes

python_on_whales/components/compose/cli_wrapper.py
MisterOwlPT: This was actually something that also came into my mind while implementing the changes.
However, the decision to require the `--yes` flag was intentional on my side for _security reasons_.

As mentioned in the [Compose official documentation](https://docs.docker.com/compose/how-tos/oci-artifact/), the Compose YAML files get pushed to a remote and there's an associated risk of exposing sensitive data when working with environment variables. Therefore, when invoking the `docker compose publish` command a safety check prompt is displayed whenever a service has an `environment` section or a global `secrets` section - this as per my experience so far.

The flag `--yes` allows users to bypass this step and is super useful for automation, but also very dangerous if used carelessly.

Therefore, by requiring users to add `--yes`, not only do we maintain the original behavior but also we are protecting users by default.

gabrieldemarmiesse: I see, i guess this option isn't used a lot in the CLI, because users can yes type "yes" when prompted to do so. In a python library this is quite unusual. And the "yes" is not very descriptive. Would that be fine by you if we rename the argument to something more descriptive and scary? Like "allow_publishing_sensitive_data"? This way people reading the code would know the weight of activating the flag. I'm scared that too many users use `yes=True` without knowing the consequences. I'd say that here, security is more important than copying the api of docker

MisterOwlPT: Honestly my concern was having the flag `--yes` set to `True` as default. 

Personally, I do not mind at all the name you propose as it is indeed much clear to the user.

If all is fine to you, I will make and commit the required changes and ask you for a new review.

Thanks for the feedback!

Edit: typos and grammar

gabrieldemarmiesse: Yes, let's change the argument name, I'll let you decide what's the best name here for the --yes of the compose commands (of course False by default)

MisterOwlPT: Hi!

As discussed I've refactored the variable name to `allow_publishing_sensitive_data`.
Here's the usage example from before adapted to the new naming:

```python
>>> from python_on_whales import DockerClient
>>> # ...
>>> docker = DockerClient(compose_files=["example_deployment.yml"])
>>> docker.compose.publish("docker.io/username/my-application:v0.0.1", allow_publishing_sensitive_data=True)
```

Let me know what you think.

python_on_whales/components/compose/cli_wrapper.py
```diff
@@ -1020,6 +1048,7 @@ def up(
         stream_logs: bool = False,
         wait_timeout: Optional[int] = None,
         dependencies: bool = True,
+        yes: bool = False,
```
gabrieldemarmiesse: I've never been asked by docker compose up to answer yes or no. How does this happen? I we don't know, maybe let's leave out this modification for now.

MisterOwlPT: When pulling Compose applications with `docker compose pull`, if a service includes environment variables, a prompt is displayed listing the inferred values for said environment variables.

I believe this is meant to help users understand how the services are about to be configured, but I am not sure this is the main reason. As before, setting `--yes` bypasses all prompts.

Even if this is seems less critical than before, I think we should keep the original behavior and instead rename the flag to something more descriptive than `yes`.

My suggestion is to rename it to something like `use_default_sensitive_values`, based on the previous comments and to better illustrate what it does. What do you think?

Edit: typos and grammar

gabrieldemarmiesse: Thanks for the renaming and sorry for the review delay. The PR seems good to me! I'll merge it.

gabrieldemarmiesse (approved):

PR #720, opened 2026-07-30
Support all buildx progress modes in build and bake

buurro:
Expands `progress` in `docker.buildx.build` and `docker.buildx.bake` to accept all modes supported by buildx: `"auto"`, `"none"`, `"plain"`, `"quiet"`, `"rawjson"`, `"tty"`.

Also rejects `stream_logs=True` with `progress="none"`/`"quiet"` in `build()`, since those modes emit no logs and the iterator would be silently empty.


gabrieldemarmiesse (approved):
Looks good to me, thanks a lot!

PR #721, opened 2026-07-30
Handle podman >= 5.0 pod inspect returning a JSON array

buurro:
Fixes an issue with newer podman versions that is causing a CI failure here: https://github.com/gabrieldemarmiesse/python-on-whales/pull/720


gabrieldemarmiesse (approved):
LGTM!

Issue #722, opened 2026-08-14
Docker bake can't handle fs_allow parameter


magyta93:
Since buildx v0.19, `docker buildx bake` refuses to run targets that need
filesystem entitlements (for example a `secret` whose `src` lives outside
the build context) and instead asks the user to confirm them interactively.
That prompt makes `bake()` hang in non-interactive Python code, and there was
no way to pass the corresponding `--allow` flag.

Add `allow` argument to `docker.buildx.bake`

PR #723, opened 2026-08-14
Add `allow` argument to `docker.buildx.bake` (#722)

magyta93:
Without it, bake targets needing entitlements (e.g. a secret whose src is outside the build context) block on an interactive confirmation prompt. Accepts a string or a list, like the existing `allow` of `build()`.

Fixes an issue here: https://github.com/gabrieldemarmiesse/python-on-whales/pull/722

### DIFF
--- a/tests/python_on_whales/components/buildx/test_buildx_cli_wrapper.py
+++ b/tests/python_on_whales/components/buildx/test_buildx_cli_wrapper.py
@@ -809,6 +809,66 @@ def test_bake_metadata_file_str_path_option(only_print, monkeypatch, tmp_path):
 
 
 @pytest.mark.usefixtures("with_docker_driver")
+@pytest.mark.usefixtures("change_cwd")
+@pytest.mark.parametrize("only_print", [True, False])
+def test_bake_without_allow_option(only_print, monkeypatch):
+    recorded = {}
+
+    def fake_run(cmd, capture_stderr=True, env={}):
+        recorded["cmd"] = list(cmd)
+        return "{}"
+
+    monkeypatch.setattr(python_on_whales.components.buildx.cli_wrapper, "run", fake_run)
+
+    docker.buildx.bake(files=[bake_file], print=only_print)
+
+    assert "--allow" not in recorded.get("cmd", [])
+
+
+@pytest.mark.usefixtures("with_docker_driver")
+@pytest.mark.usefixtures("change_cwd")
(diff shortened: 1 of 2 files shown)


tests/python_on_whales/components/buildx/test_buildx_cli_wrapper.py
```diff
@@ -808,6 +808,66 @@ def fake_run(cmd, capture_stderr=True, env={}):
...
+
+    cmd = list(map(str, recorded.get("cmd", [])))
+    assert cmd.count("--allow") == 1
+    assert cmd[cmd.index("--allow") + 1] == "network.host"
+
+
+@pytest.mark.usefixtures("with_docker_driver")
+@pytest.mark.usefixtures("change_cwd")
+@pytest.mark.parametrize("only_print", [True, False])
+def test_bake_allow_multiple_options(only_print, monkeypatch):
```
gabrieldemarmiesse, line 850: can we try to add one test without monkeypatching?

gabrieldemarmiesse (approved):
Thanks for the PR, yeah that makes sense. I think we just need one more test without monkeypatching to prove that the code works well with the actual docker client, and then we should be good to go!

tests/python_on_whales/components/buildx/test_buildx_cli_wrapper.py
magyta93: Sure! I added three tests that run against the actual docker client, no monkeypatching:

test_bake_allow_with_real_docker_client — runs a real docker buildx bake with a single entitlement and with a list of them, for both print=True and print=False, and checks the returned config. This proves the CLI accepts the flags we generate.
test_bake_invalid_allow_with_real_docker_client — passing an unknown entitlement makes docker fail with unknown entitlement key, which proves the value really reaches the CLI instead of being silently dropped.
test_bake_allow_grants_entitlement — the functional one: a bake target whose secret lives outside the build context (the use case that motivated this PR). Without allow buildx refuses with additional privileges requested, and with allow="fs.read=<dir>" the build succeeds and the image is created. It skips itself if the installed buildx doesn't enforce filesystem entitlements, so it won't be flaky on older versions.
I kept the monkeypatched ones as well since they pin down the exact command line (one --allow per entitlement, and no --allow at all when the argument isn't used), which isn't observable from the outside. Happy to drop them if you'd rather not have mocks in there.

All of them pass locally against docker.

PR #724, opened 2026-08-19
feat: support buildx build resource limits

marxarelli:
Add support for the `--resource` buildx build flag added in buildx 0.29 and supported by BuildKit >= 0.31.

### DIFF
--- a/python_on_whales/components/buildx/cli_wrapper.py
+++ b/python_on_whales/components/buildx/cli_wrapper.py
@@ -255,6 +255,7 @@ class BuildxCLI(DockerCLICaller):
         provenance: Union[bool, Dict[str, str], None] = None,
         pull: bool = False,
         push: bool = False,
+        resource: Optional[Dict[str, str]] = None,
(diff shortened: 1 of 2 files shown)


marxarelli:
@gabrieldemarmiesse looking for a reviewer.

Unfortunately I could not enable the test due to the older Docker Engine version in ubuntu-24.04 and the older version of buildx installed in CI.

I'm happy to work on upgrading the CI configuration if that's helpful, although I did encounter some additional failures when running the test suite against a newer buildx (FWICT bake metadata output changed slightly at some point).

There are some CI failures that do not seem related to my change.

gabrieldemarmiesse:
@marxarelli  I'm very sorry for the delay in the review process. Thanks a lot for the PR, this features makes a lot of sense! Yes feel free to upgrade the docker version in the CI so that we can test that. It's usually better to have the most recent version of docker in the CI, since we can test more stuff with it. Of course if the version can be pinned it's best to avoid sudden breakages when a new version is out.

